PR #25, opened 2019-08-06
Summary

### LINKED ISSUE #12: summarize the raw results
When performing a comparison, for each file, a list of matching files is returned. For each of those matching files, a list of matching lines is returned.

- At the lowest level, the list of matches should be converted to a list of ranges
- A similarity score per pair of files should be calculated (based on the number of matching lines or the number of matching k-mers)
- When reporting, there are 2 options:
  - Pick a file and get the list of matching files, sorted from highest to lowest similarity
  - Show a list of the top pairwise comparisons (this is what moss does). While ok, I’m not a great fan of this because if there are bigger groups of cooperating students, these will take up a large portion at the top of the result table. Such groups should actually be grouped and not listed pair-wise.

ISteampowered:
closes #12


src/lib/summary.ts
```diff
@@ -0,0 +1,330 @@
...
+      }
+    }
+  }
+
+  /**
+   * attempts to extend one range with the other. If it fails then it returns undefined.
+   * @param range1 the first range you want to extend
+   * @param range2 the second range you want to extend
+   */
+  public extendRangeWithRange(range1: Range, range2: Range): Range | undefined {
```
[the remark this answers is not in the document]
ISteampowered: No there is not, apart from that there needs to be a test to see if the two ranges can extend each other.

src/app.ts
```diff
@@ -1,10 +1,27 @@
...
+  // }
+  // const result = await comparison.compareFiles(arr);
+
+  await comparison.addFiles(["samples/js/sample.js"]);
+  const result = await comparison.compareFiles([
+    "samples/js/copied_function.js",
+    "samples/js/another_copied_function.js",
+  ]);
+  const summary = new Summary(result);
+  console.log(summary.toString());
```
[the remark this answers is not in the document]
ISteampowered: This does not work. If you remove the toString method call then the object itself is shown and the toString is not called.

src/lib/summary.ts
```diff
@@ -0,0 +1,269 @@
...
+    // If two rangesTuples overlap with each other then extend the second with the first and remove the
+    // first from the array.
+    for (let i = ranges.length - 1; i >= 0; i--) {
+      for (let j = i; j >= 0; j--) {
+        if (i !== j) {
+          if (this.canExtentRangesTupleWithRangesTuple(ranges[i], ranges[j])) {
+            this.extendRangesTupleWithRangesTuple(ranges[j], ranges[i]);
+            ranges.splice(i, 1);
+          }
+          break;
```
bmesuere: I can be mistaken, but shouldn't this break be in the if? If you extended the range, the i'th element is removed, so the outer loop should start a new iteration with a new i.

ISteampowered: You are correct, I fixed this in the feature/CLI-2 branch and added a test for it but forgot to make the same changes in this branch.

src/lib/range.ts
```diff
@@ -10,16 +10,16 @@ export class Range {
     this.to = to;
   }
 
-  public getFrom(): number {
+  public get From(): number {
```
bmesuere: use lower case property fields

Issue #27, opened 2019-08-06
Compare with MOSS results


ISteampowered:
[MOSS](https://theory.stanford.edu/~aiken/moss/)

ISteampowered:
at the time of commit 7a6fd96 all the dolos flags match the moss flags. So from now on the same flags for both applications are used.

ISteampowered:
This compares the -d flag.
Both application are run with `-l javascript -m 10 -d`. The structure of the dummy files are:
```
assignment1/
├── student1
│   ├── main.js
│   ├── sample.js
│   └── subDirectory
│       └── childClass.js
├── student2
│   ├── copied_function.js
│   ├── helperClasses
│   │   └── childClass.js
│   └── main.js
└── student3
    ├── another_copied_function.js
    ├── main.js
    └── tempName
        ├── childClass.js
        ├── hello.js
        └── subDir
            └── subsubClass.js
```
Where everything except `sample.js`, `copied_function.js` and `another_copied_function.js` is empty. Both `copied_function.js` and `another_copied_function.js` contain lines from `sample.js`
Here the results first three results from moss have a corresponding match within the dolos results. After that the moss results become very odd which I can only assume is a bug of sorts. Another noticable difference lies with the sorting, but this is intentional (descibed in #12 ).

ISteampowered:
Both programs were run with `-l javascript -m 10`. When comparing 51 real exercises from student then moss takes ~11s while dolos takes ~26s. The output of dolos can't really be compared as the output of this contains 31875 lines.

bmesuere:
Can you post an updated runtime with the new code? Be sure to only run the code and not to include the tsc compile step.

ISteampowered:
Yeah sure, at the time of commit e7c8b14 running dolos with the same files as in my previous comment takes roughly 2s.

ISteampowered:
It seems that our algorithm can't handle the following code segment very well: 
```js
        let minuten = {
            "00": "HET IS",
            "05": "HET IS VIJF OVER",
            "10": "HET IS TIEN OVER",
            "15": "HET IS KWART OVER",
            "20": "HET IS TIEN VOOR HALF",
            "25": "HET IS VIJF VOOR HALF",
            "30": "HET IS HALF",
            "35": "HET IS VIJF OVER HALF",
            "40": "HET IS TIEN OVER HALF",
            "45": "HET IS KWART VOOR",
            "50": "HET IS TIEN VOOR",
[...]

PR #32, opened 2019-08-14
CLI

### LINKED ISSUE #11: filter the comparison output
The output should be filtered in at least 3 ways:
- exclude matches with the file itself (at post-processor)
- matches with given boilerplate code (add a blacklist to `Comparison`)
- common hashes matching many files. These are probably common constructs. The threshold for this could be user-defined. (at post-processor or when comparing)

### LINKED ISSUE #14: make dolos available as a CLI
Next to the current library, also offer Dolos as a CLI (and publish it through npm).

ISteampowered:
closes #14  and closes #11


tsconfig.json
```diff
@@ -60,7 +60,8 @@
     // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
     // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
     "pretty": true,
-    "forceConsistentCasingInFileNames": true
+    "forceConsistentCasingInFileNames": true,
+    "resolveJsonModule": true,
```
bmesuere: This is the reason why your build location has changed. As long as all source files are in a single directory (i.e. src), tsc will leave out that directory in the dist directory (which is what we want).

You included this option to be able to require the package.json file and fetch the version from there. I would remove this option and read the package file in another way or, if impossible, duplicate the version number or leave out the version command.

src/command.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    // Compare the base file with all the other files when there is a base file.
+    if (program.base) {
+      const baseFileComparison = new Comparison(tokenizer);
+      await baseFileComparison.addFile(program.base);
+      baseFileMatches = await baseFileComparison.compareFiles(locations);
+    }
+
+    // Compare all the file with each other.
+    while (locations.length > 1) {
```
bmesuere: This way, you're building the index for each file which is absolutely unnecessary. You only need to create it once.
- create a new comparison object
- add all files to the index (we have a utility method for this to do it for all files at once)
- for each of the files, query the index (we also have a utility function for this)

There is no need to aggregate the results yourself since this is already done by the comparison object.

src/lib/summary.ts
```diff
@@ -1,52 +1,71 @@
 import fs from "fs";
 import { Matches } from "./comparison";
 import { Range } from "./range";
+import { SummaryFilter } from "./summaryFilter";
 export type RangesTuple = [Range, Range];
+
 export class Summary {
+  public readonly gapSize: number;
+  public readonly summaryFilter: SummaryFilter;
+  public readonly comment: string | undefined;
```
bmesuere: This should not be in the summary

src/lib/summary.ts
```diff
@@ -1,52 +1,71 @@
...
+   * Generates a summary for the given matches.
    * @param matches A many-to-many comparison of a set of files. This map contains an entry for each of the
    * input files with the key being its file name and the value a list of matches. These matches are grouped
    * per matching file. The compareFiles function of the Comparison class can generate such mapping.
-   * @param minimumMaximumLines The minimum amount of lines required by the longest range in a rangesTuple. If the
-   * rangesTuple has less lines then it will be filtered out. When the rangesTuple has two ranges with a different
-   * amount of lines, then the maximum between to two is used.
-   * @param minimumMinimumLines The minimum amount of lines required by the shortest range in a rangesTuple.
+   * @param summaryFilter The summary filter that will be used to filter the results.
    * @param gapSize The gap size allowed during the joining of two ranges. For example if the gap size is 0 then [1,3]
```
bmesuere: some parameters are missing

src/lib/summaryFilter.ts
```diff
@@ -0,0 +1,333 @@
+import { Matches } from "./comparison";
+import { Range } from "./range";
+import { RangesTuple } from "./summary";
+
+export class SummaryFilter {
```
bmesuere: We should discuss the use of this filter object in person. The current implementation is not user-friendly because the "user" has to create this object as well as a summary object with the options split between them.

src/lib/summaryFilter.ts
```diff
@@ -0,0 +1,333 @@
...
+  /**
+   * Filter the tuples pair based on the maches with the base file.
+   * @param matchingLinesArray A matching line array between two files that you want to be filtered.
+   * @param baseFileMatches1 The base file matches with the first file. The first elements of each tuples in this will
+   * be used to filter the tuples based on the first element of each tuple in [[matchingLinesArray]].
+   * @param baseFileMatches2 The same as [[baseFileMatches1]] but then for the second file. Here the first elements
+   * in the tuples from this array will be used to filter the second element from the tuples from
+   * [[matchingLinesArray]].
+   */
+  public static filterByBaseFile(
```
bmesuere: base file stuff should be in the comparator

src/command.ts
```diff
@@ -0,0 +1,236 @@
...
+  )
+  .option(
+    "-M --maximum-percentage <float>",
+    "Maximum percentage a passage may appear before it is ignored. The percentage is calculated using the amount of " +
+      "different groups there are. So with the -d options the amount of directories is used where normally the " +
+      "amount of files is used. Must be a value between 1 and 0.",
+    0.9,
+  )
+  .option("-c, --comment <string>", "comment string that is attached to the generated report")
+  .option(
```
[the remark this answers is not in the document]
ISteampowered: By default there is nothing added to the report.

src/command.ts
```diff
@@ -0,0 +1,236 @@
...
+    "-n, --file-amount",
+    "The -n option specifies how many matching file are shown in the result",
```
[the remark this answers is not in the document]
ISteampowered: It should be pairs.

src/command.ts
```diff
@@ -0,0 +1,236 @@
...
+    0,
+  )
+  .option(
+    "-z, --zero-based-lines",
+    "Specifies whether or not you want lines to be zero based",
+    false,
+  )
+  .arguments("<locations...>")
+  .action(filesArgs => {
+    locations = filesArgs;
```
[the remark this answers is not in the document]
ISteampowered: I think commander calls the given function when it is appropriate to do so. This is how it was shown in the examples.

src/command.ts
```diff
@@ -0,0 +1,236 @@
...
+  let groupAmount: number;
+  const tokenizer = new CodeTokenizer(program.language);
+  const results: Map<string, Matches<number>> = new Map();
+  let baseFileMatches: Map<string, Matches<number>> = new Map();
+
+  // if the -d and the -b flag are active, filter out all the basefile locations
+  if (program.directory && program.base) {
+    program.base = path.normalize(program.base);
+
+    baseFiles = locations.filter(filePath => filePath.startsWith(program.base));
```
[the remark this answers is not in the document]
ISteampowered: No there is not, but when both the -d and the -b options are used at the same time, the -b options only specifies the directory where the base files are located. The actual files themself are given with the rest of the files. I did it it this way because commander cannot handle two variable arguments at the same time.

src/command.ts
```diff
@@ -0,0 +1,236 @@
...
+const program = new commander.Command();
+
+// Initial program description and version
+program.version(packageJson.version).description("Plagiarism detection for programming exercises");
+
+let locations: string[] = [];
+
+program
+  .option("-l, --language <language>", "Language used in the compared programs.", "javascript")
+  .option("-d, --directory", "Specifies that submission are per directory, not by file.")
```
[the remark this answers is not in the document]
ISteampowered: It is true that the current implementation is not efficient, this was pointed out [in this comment](https://github.com/dodona-edu/dolos/issues/27#issuecomment-522562238). I simply tried to match the arguments from the moss application as much as possible.

ISteampowered:
I have noticed that there are some strange results when comparing files that I have to fix first.

src/lib/comparison.ts
```diff
@@ -21,10 +45,18 @@ export class Comparison<Location> {
    * @param tokenizer A tokenizer for the correct programming language
    * @param hashFilter An optional HashFilter to filter the hashes returned by
```
bmesuere: Is this still used? it's no longer a parameter.

src/lib/comparison.ts
```diff
@@ -21,10 +45,18 @@ export class Comparison<Location> {
...
    * the rolling hash function.
+   * @param noFilter A NoFilter used to generate hashes for blacklisted files.
+   * @param passageOptions The options used to filter based on the passage count. For a more detailed explanation see
+   * [[ComparisonOptions]]. If this options is not used then no filtering will occur.
    */
-  constructor(tokenizer: Tokenizer<Location>, hashFilter?: HashFilter) {
+  constructor(tokenizer: Tokenizer<Location>, options?: ComparisonOptions) {
```
bmesuere: if options only contains 3 possible values, use object destructuring in the argument list.

src/app.ts
```diff
@@ -1,26 +1,124 @@
...
+    filterPassageByPercentage: program.maximumPassageCount === undefined,
+    maxPassage:
+      program.maximumPassageCount === undefined
+        ? program.maximumPassagePercentage
+        : program.maximumPassageCount,
+  };
+
+  // Compare all the file with each other.
+  const comparison = new Comparison(tokenizer, {
+    filterOptions: comparisonPassageFilterOptions,
```
bmesuere: why are these in a separate subobject and not in the "parent" options object?

src/lib/comparison.ts
```diff
@@ -98,19 +131,58 @@ export class Comparison<Location> {
...
-      const matches = this.index.get(hash);
-      if (matches) {
-        for (const [fileName, lineNumber] of matches) {
-          const match: [Location, Location] = [lineNumber, mapping[location]];
-          const lines = matchingFiles.get(fileName);
-          if (lines) {
-            lines.push(match);
-          } else {
-            matchingFiles.set(fileName, [match]);
+      if (!this.filteredHashSet.has(hash)) {
```
bmesuere: use `if (this.filteredHashSet.has(hash)) { continue; }` as a guard clause instead of this nested block

src/lib/comparison.ts
```diff
@@ -98,19 +131,58 @@ export class Comparison<Location> {
...
+        const matches = this.index.get(hash);
+        if (matches && this.filterByPassageCount(matches.length)) {
+          for (const [fileName, lineNumber] of matches) {
+            // add the match if the match is not with the file and if the file comes first when alphabetically sorted.
+            // This is done to avoid the duplicates in the following case: when file A matches with file B, file B will
```
bmesuere: Our similarities are not symmetric, so I think we want to compare both A with B and B with A.
Example, file A contains 10 functions, file B contains only a single function which was copied from A. The similarities here are 10% in one direction and 100% in the other.

src/lib/summary.ts
```diff
@@ -2,51 +2,126 @@ import fs from "fs";
...
+ */
+export interface FilterOptions {
+  passageOutputLimit?: number;
+  minimumLinesInLargestPassage?: number;
+  minimumLinesInSmallestPassage?: number;
+}
+
 export class Summary {
+  public readonly gapSize: number;
+  public readonly comment: string | undefined;
```
bmesuere: I still don't think the comment has to be passed onto the summary. The comment is something output related and has no place here.

src/app.ts
```diff
@@ -1,26 +1,120 @@
...
+      "of code that appears in many programs is probably legitimate sharing and not the result of plagiarism. With " +
+      "-M N any fragment appearing in more than N percent of the files is filtered out. " +
+      "Must be a value between 0 and 1.",
+    0.9,
+  )
+  .option("-c, --comment <string>", "Comment string that is attached to the generated report")
+  .option(
+    "-n, --fragment-output-limit",
+    "Specifies how many matching fragment are shown in the result. All fragment are " +
+      "shown then this option isn't used.",
```
bmesuere:
```suggestion
      "shown then this option is omitted.",
```

src/app.ts
```diff
@@ -1,26 +1,120 @@
...
+    "The -M option sets how many percent of the files the code fragment may appear before it is ignored. A fragment " +
+      "of code that appears in many programs is probably legitimate sharing and not the result of plagiarism. With " +
+      "-M N any fragment appearing in more than N percent of the files is filtered out. " +
+      "Must be a value between 0 and 1.",
+    0.9,
+  )
+  .option("-c, --comment <string>", "Comment string that is attached to the generated report")
+  .option(
+    "-n, --fragment-output-limit",
+    "Specifies how many matching fragment are shown in the result. All fragment are " +
```
bmesuere: shouldn't this be code file pairs ("matching pairs") instead of fragments?

src/app.ts
```diff
@@ -1,26 +1,120 @@
...
+      "shown then this option isn't used.",
+  )
+  .option(
+    "-s, --minimum-lines-shortest <integer>",
+    "The minimum amount of lines in the shortest code fragment in a comparison before it is shown",
```
bmesuere: I still don't get the -s and -S options, especially the "shortest" bit. Can you explain this to me with an example?

src/app.ts
```diff
@@ -1,26 +1,120 @@
...
+    1,
+  )
+  .option(
+    "-S, --minimum-lines-longest <integer>",
+    "The minimum amount of lines in the longest code fragment in a comparison before it is shown",
+    1,
+  )
+  .option(
+    "-g, --maximum-gap-size <integer>",
+    "If two fragment need to be joined, then this parameter specifies how large the gap between the two fragment may" +
```
bmesuere: "If two fragments are close to each other, they will be merged into a single fragment if the gap between them is smaller than the given number of lines."

src/lib/comparison.ts
```diff
@@ -19,12 +49,14 @@ export class Comparison<Location> {
    * After creation, first add files to the index which can then be queried.
    *
    * @param tokenizer A tokenizer for the correct programming language
-   * @param hashFilter An optional HashFilter to filter the hashes returned by
-   * the rolling hash function.
+   * @param fragmentOptions The options used to filter based on the fragment count. For a more detailed explanation see
+   * [[ComparisonOptions]]. If this options is not used then no filtering will occur.
    */
-  constructor(tokenizer: Tokenizer<Location>, hashFilter?: HashFilter) {
+  constructor({ tokenizer, hashFilter, noFilter, filterOptions }: ComparisonOptions<Location>) {
```
bmesuere: Sorry, my previous comment probably wasn't clear. I didn't want you to bring the tokenizer inside the options hash, but bring the maxFragment and filterFragmentByPercentage to a lower level (and get rid of filter options).

src/lib/summary.ts
```diff
@@ -2,51 +2,127 @@ import fs from "fs";
...
+
 export class Summary {
+  public readonly gapSize: number;
   private readonly results: Map<string, Matches<Range>>;
-  private readonly minimumMaximumLines: number;
-  private readonly gapSize: number;
-  private readonly minimumMinimumLines: number;
+  private readonly filterOptions: FilterOptions;
+
+  private readonly defaultFilterOptions: FilterOptions = {
```
bmesuere: If these values aren't meant to be changed, also wrap the right part with `Object.freeze()`

bmesuere: If you make this static, you can use it as a default value in your constructor instead of manually checking inside of it.

src/app.ts
ISteampowered: When we have a pair of ranges like for example [[3, 5], [6, 11]]. The shortest of these two is the first one, we can filter the minimum amount of lines with the -s option. The minimum amount of lines in the second one can be set with the -S option.

bmesuere: I'm afraid I still don't get it. I agree we need a "minimal fragment size" option: we don't display ranges shorter than that (we can discuss if both or the ranges of a couple must meet the requirement or only one).

But why would you want to filter out long fragments?

bmesuere: Ok, sorry, now I get it, both are about a minimal fragment size, but you can set a different value for both entries in a couple.

I would simplify this to "minimal fragment length" and only have one option.

src/lib/comparison.ts
```diff
@@ -98,19 +131,62 @@ export class Comparison<Location> {
...
+    }
+  }
+
+  /**
+   *
+   * @param fragmentCount The amount of times a certain code fragment has appeared
+   * @returns true if is value is acceptable under the options given in the contructor or when the options are not
+   * defined.
+   */
+  private filterByFragmentCount(fragmentCount: number): boolean {
```
bmesuere: This naming can be confusing. It used to be a fragment count filter, but right now it is used as a hash count filter. (The name fragment is used as a synonym for range to the end-user)

src/app.ts
```diff
@@ -1,26 +1,116 @@
...
+  // Compare the base file with all the other files when there is a base file.
+  if (program.base) {
+    comparison.addFileToFilterList(program.base);
+  }
+
+  await comparison.addFiles(locations);
+  const matchesPerFile: Map<string, Matches<number>> = await comparison.compareFiles(locations);
+
+  const filterOptions: FilterOptions = {
+    fragmentOutputLimit: program.filePairOutputLimit,
```
bmesuere: This was renamed to filePairOutputLimit, but it seems the behaviour stayed the same? If I'm correct this limits the number of fragments that are reported per file pair?

What is should be is the number of file pairs that are reported. For example, when you have 100 files/users, you have 10000 filepairs. You only want to report on the top n pairs.

ISteampowered: Ah sorry I misunderstood you last time. I will change the implementation.

PR #37, opened 2019-08-28
JSON and html formatters

ISteampowered:


src/app.ts
```diff
@@ -53,6 +54,16 @@ program
...
   )
+  .option(
+    "-o, --output-format <format>",
+    "Specifies what format the output should be, current options are: terminal/console, json, html.",
+    "terminal",
+  )
+  .option(
+    "-v, --cluster-cut-off-value <number>",
+    "The minimum amount of lines needed before two files will be clustered together",
+    13,
```
bmesuere: wouldn't it be better to use a percentage score (i.e. the score value) here so it doesn't depend on file size?

src/lib/assets/scripts.js
```diff
@@ -0,0 +1,7 @@
+
+function show(shown, hidden) {
```
bmesuere: - I think it's cleaner to use 2 functions for this (show/hide) or rename it to swap or something
- the console log must be removed
- instead of setting display block, `initial` is probably a better value

src/lib/range.ts
```diff
@@ -32,6 +32,10 @@ export class Range {
     return this.from - 1 - gapSize <= value && value <= this.to + 1 + gapSize;
   }
 
+  public isNumberWithingBounds(value: number): boolean {
```
bmesuere: rename to `includes`

src/app.ts
ISteampowered: I though about that but I think doing it my way would be slightly more clear and allows the users to more easily pick a value that is relevant for them.

rien:
Merged into develop with 174c262

PR #44, opened 2019-09-02
implemented -d option

ISteampowered:
depends on #37


rien:
Merged into develop with d95822a460b88b5a8faff9423941eab9b10bfffd

PR #81, opened 2019-11-06
benchmark dataset

yasminosb:


samples/python/benchmark_files/UPGMA_A_functionsmoved.py
```diff
@@ -0,0 +1,243 @@
+from collections import defaultdict
```
rien: Which functions are moved?

samples/python/benchmark_files/UPGMA_A_variablenames.py
```diff
@@ -0,0 +1,243 @@
+from collections import defaultdict
```
rien: Which variable names have been changed?

PR #115, opened 2019-12-11
cli cleanup

yasminosb:
all indentation should be flawless now, no more trailing newlines caused by the default paramaters that were not considered yet

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -139,7 +139,7 @@ program
         maxGapSize: program.maxGapSize,
         maxHashCount: program.maximumHashCount,
         maxHashPercent: program.maximumHashPercentage,
-        maxMatches: program.filePairOutputLimit, // still needed?
+        maxMatches: program.filePairOutputLimit,
         minFragmentLength: program.minimumFragmentLength,
       });
       const matches = await dolos.analyze(locations);
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -8,14 +8,22 @@ import * as Utils from "./lib/utils";
 const pkg = require("../package.json");
 const program = new Command();
 
-const indentLength: number = 43;
+const indentLength: number = 42;
 const maxLineLength: number = (process.stdout.columns as number) - indentLength;
+const defaultLength: number = 12; // " (default: )"
 
-function indent(helpText: string): string {
+/**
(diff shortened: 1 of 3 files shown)


src/lib/options.ts
```diff
@@ -9,7 +9,7 @@ export interface CustomOptions {
   maxGapSize?: number;
   maxHashCount?: number;
   maxHashPercent?: number;
-  maxMatches?: number;
+  maxMatches?: number; // still needed?
```
rien:
```suggestion
  maxMatches?: number;
```

The `maxMatches` option is still used [here](https://github.com/dodona-edu/dolos/blob/1d6061497c34918555f7e4b02b54f9a53c3bb76a/src/lib/summary.ts#L89), or what do you mean?

src/cli.ts
```diff
@@ -125,7 +139,7 @@ program
         maxGapSize: program.maxGapSize,
         maxHashCount: program.maximumHashCount,
         maxHashPercent: program.maximumHashPercentage,
-        maxMatches: program.filePairOutputLimit,
+        maxMatches: program.filePairOutputLimit, // still needed?
```
rien:
```suggestion
        maxMatches: program.filePairOutputLimit,
```

src/lib/options.ts
```diff
@@ -89,7 +89,7 @@ export class Options {
     return definedOrDefault(this.custom.clusterMinMatches, Options.defaultClusterMinMatches);
   }
 
-  get maxMatches(): number | null {
+  get maxMatches(): number | null { // still needed?
```
rien:
```suggestion
  get maxMatches(): number | null {
```

src/lib/options.ts
```diff
@@ -105,7 +105,7 @@ export class Options {
       maxGapSize: this.maxGapSize,
       maxHashCount: this.custom.maxHashCount,
       maxHashPercent: this.filterByPercentage ? this.maxHashPercent : undefined,
-      maxMatches: this.maxMatches,
+      maxMatches: this.maxMatches, // still needed?
```
rien:
```suggestion
      maxMatches: this.maxMatches,
```

src/lib/options.ts
yasminosb: I must have overlooked that reference! I removed the comments.

rien:
Merged manually with aed2aa1

PR #122, opened 2019-12-13
Update tree sitter to 0.16.x

rien:
This uses the updated tree-sitter packages which uses a new binding API which allows us to use newer versions of Node (> 10).


rien:
The C-sharp grammar is currently broken because its `index.js` points to an old location of the binding file. A PR has been created fixing this issue (https://github.com/tree-sitter/tree-sitter-c-sharp/pull/64).

Once the C-sharp grammar has been fixed and the tests are working, we can merge this PR.

PR #255, opened 2020-08-03
Rename classes

### LINKED ISSUE #233: Rename some classes
One of the hardest things in computer science.

We have the following classes (from bottom to top). I think names **in bold** need another name, but suggestions for the other classes are welcome as well.
- `File`: a submission with source code belonging to a student.
- `SharedKmer`: a substring of the AST which is shared by at least two files (which keeps a reference to those files).
- **`Match`**: a connector between two matching files. It keeps a reference to the `SharedKmer` and the location within those two files.
- **`Selection`**: a part of a file which has a match.
- **`Fragment`**: a collection of one or more subsequent `Match`es. These will be visualized as one contiguous "block" of code which is shared between two files.
- **`Intersection`**: a comparison between two files, collecting all the `Fragment`s shared between them. The name is chosen because it contains the "overlap" in code between the files.
- `Analysis`: which processes and manages the result of analyzing a collection of files (submissions), essentially a list of `Intersections`.
- **`Comparison`**: the class which actually compares one file with the _index_ of kmers. It takes a list of `File`s and returns an `Analysis`.

rien:
FilePart -> Occurence
Match -> PairedOccurrence
Fragment -> Block
Intersection -> Diff
Comparison -> Index
Analysis -> Report

I think the `FilePart` needs to be refactored a bit to contain a Region, File and Kmer. So FilePart/Occurence and Side/ASTRegion should be merged.
Closes #233


pdawyndt:
OccurencePair -> PairedOccurrence?

pdawyndt:
Alternative suggestion for hunk/fragment: clone. The name itself already suggests copying in two or more places.

rien:
We've noticed that we've almost automatically started describing the hunks as (duplicated) blocks.

pdawyndt:
Rename into block, clone, <something else>?

ISteampowered:
TODO: 
- [ ] cli:
    - [ ] FilePart -> Occurence
    - [ ] Match -> PairedOccurrence
    - [ ] Fragment -> Block
    - [ ] Hunk -> Block
    - [ ] Intersection -> Diff
    - [ ] Comparison -> Index 
    - [ ] Analysis -> Report
- [ ] web UI
    - [ ] FilePart -> Occurence
    - [ ] Match -> PairedOccurrence
    - [ ] Hunk -> Block
    - [ ] Fragment -> Block
    - [ ] Intersection -> Diff
    - [ ] Comparison -> Index 
    - [ ] Analysis -> Report
    - [ ] check csv input will no longer work after cli rename fix this

ISteampowered:
Closes #233

cli/src/cli.ts
```diff
@@ -74,7 +74,7 @@ program
   .option(
     "-c --compare",
     Utils.indent(
-      "Print a comparison of the matches even if analysing more than two " +
+      "Print a comparison of the pairedOccurrences even if analysing more than two " +
```
rien, line 77:
```suggestion
      "Print a comparison of the matching blocks even if analysing more than two " +
```

cli/src/cli.ts
```diff
@@ -44,7 +44,7 @@ program
   )
   .option(
     "-d, --directory",
-    Utils.indent("Specifies that submision are per directory, not by file. ")
+    Utils.indent("Specifies that submission are per directory, not by file. ")
```
rien, line 47:
```suggestion
    Utils.indent("Specifies that submissions are per directory, not by file. ")
```

cli/src/cli.ts
```diff
@@ -91,13 +91,13 @@ program
     x => parseFloat(x)
   )
   .option(
-    "-s, --minimum-fragment-length <integer>",
+    "-s, --minimum-block-length <integer>",
     Utils.indent(
-      "The minimum length of a fragment. Every fragment shorter than this is " +
+      "The minimum length of a block. Every block shorter than this is " +
```
rien, line 96:
```suggestion
      "The minimum length of a matching block. Every block shorter than this is " +
```

cli/src/cli.ts
```diff
@@ -111,8 +111,8 @@ program
   .option(
     "-g, --maximum-gap-size <integer>",
```
rien, line 112: This option can be removed in #263

cli/src/lib/analyze/block.ts
```diff
@@ -1,27 +1,27 @@
...
+import { Region } from "../util/region";
 import { Range } from "../util/range";
 
 /**
- * A fragment is a collection of one or more consequent matches (kmers).
+ * A block is a collection of one or more consequent pairedOccurrences (kmers).
  *
- * A fragment can be extended with a new match if its kmer indices in both
- * files are directly after that of the fragment.
+ * A block can be extended with a new match if its kmer indices in both
```
rien, line 9:
```suggestion
 * A block can be extended with a new PairedOccurence if its kmer indices in both
```

cli/src/lib/analyze/block.ts
```diff
@@ -1,27 +1,27 @@
...
+ * files are directly after that of the block.
  */
-export class Fragment {
+export class Block {
 
-  public matches: Array<Match>;
+  public pairedOccurrences: Array<PairedOccurrence>;
```
rien, line 14: I would use shorter variable and function names for `PairedOccurence`s. Either `occurrence` or `pair` maybe? What makes most sense for you?

cli/src/lib/analyze/index.ts
```diff
@@ -129,14 +129,14 @@ export class Comparison {
    *
    * @param file The file to query
    * @param hashFilter An optional HashFilter. By default the HashFilter of the
-   * Comparison object will be used.
-   * @return a promise of a list of Intersection objects. An Intersection object
-   * represents the common hashes (matches) between two files.
+   * Index object will be used.
+   * @return a promise of a list of Diff objects. An Diff object
```
rien, line 133:
```suggestion
   * @return a promise of a list of Diff objects. A Diff object
```

cli/src/lib/analyze/index.ts
```diff
@@ -129,14 +129,14 @@ export class Comparison {
...
+   * represents the common hashes (pairedOccurrences) between two files.
```
rien, line 134:
```suggestion
   * contains the common hashes (occurrences) between two files.
```

cli/src/lib/analyze/block.ts
ISteampowered: I think that pair will make the most sense in a lot of cases.

PR #260, opened 2020-08-07
Add proper analysis data more closely resembling our actual use-case

rien:
Loading these files is currently performed on the main thread and fairly slow, thus hanging the site. In the future we'll want to look at using serviceworkers to perform this task in the background.


rien:
Because this PR has diverged and we switched over to testing with the SOCO dataset, this PR will be closed.

PR #262, opened 2020-08-12
Feature/barcode chart

ISteampowered:
depends on #256


web/src/api/api.ts
```diff
@@ -104,22 +103,33 @@ async function fetchMetadata(
   return await d3.csv(url);
 }
 
+async function _fetchBlocks(
```
rien, line 106:
```suggestion
async function fetchBlocks(
```

cli/src/lib/presenter/csvPresenter.ts
```diff
@@ -28,6 +29,40 @@ export function writeCSVto<T>(
...
+              index: pair.right.index,
+            }
+          };
+        })
+      };
+    }), null, 2);
+  }
+
+  public writeBlocks(out: Writable, diffId: number): void {
+    const diffs = this.report.scoredDiffs.filter(s => s.diff.id === diffId);
```
rien, line 58: I would directly pass the `diff` here instead of its ID.

cli/src/lib/presenter/webPresenter.ts
```diff
@@ -23,6 +23,11 @@ export class WebPresenter extends CsvPresenter {
     app.get("/data/files.csv", respondwithCSV(o => this.writeFiles(o)));
     app.get("/data/diffs.csv", respondwithCSV(o => this.writeDiffs(o)));
     app.get("/data/kmers.csv", respondwithCSV(o => this.writeKmers(o)));
+    app.get("/data/blocks/:id.json", ((req, res) => {
```
rien, line 26: We'll probably want to return a 404 if the `id` is incorrect.

I suggest creating a mapping between the `Diff`'s and their id's in the web prestenter, so you make this lookup more easily.

web/src/api/api.ts
```diff
@@ -170,6 +179,11 @@ function parseMetadata(data: d3.DSVRowArray): Metadata {
   );
 }
 
+export async function fetchBlocks(diffId: number, kmers: ObjMap<Kmer>): Promise<Array<Block>> {
```
rien, line 182: We'll probably want to keep `Diff.blocks`, but initialize it with `null`. This function could then populate the blocks in the given `Diff`.  This function could then be named `populateBlocks`.

web/src/components/BarcodeChart.vue
```diff
@@ -0,0 +1,203 @@
...
+  @Prop() amountOfLinesVisible!: number;
+  @Prop() selectedSelections!: Array<string>;
+  @Prop() hoveringSelections!: Array<string>;
+
+  chart: {
+    width: number;
+    height: number;
+  } = {
+    width: 40,
+    height: 1000,
```
rien, line 52: We don't want to hardcode this. I would let this component resize using CSS and bind the width and height of the chart.

web/src/components/BarcodeChart.vue
```diff
@@ -0,0 +1,203 @@
...
+      .attr("height", function (d) { return y(d[0]) - y(d[1]); })
+      .attr("width", this.chart.width)
+      .on("mouseover", (a, b, rect) => this.mouseover(map, a, b, rect))
+      .on("mouseleave", (a, b, rect) => this.mouseleave(map, a, b, rect))
+      .on("click", (a, b, rect) => this.mouseclick(map, a, b, rect));
+  }
+
+  getKey(rect: SVGRectElement[] | ArrayLike<SVGRectElement>): number {
+    // @ts-expect-error
+    return d3.select(rect[0].parentNode).datum().key;
```
rien, line 142: Which error do you expect here?

web/src/components/CompareCard.vue
```diff
@@ -1,46 +1,344 @@
...
 <script lang="ts">
+
 import { Component, Vue, Prop } from "vue-property-decorator";
-import { Diff } from "@/api/api";
+import { Block, Diff, Selection } from "@/api/api";
+import CompareSide from "@/components/CompareSide.vue";
+import BarcodeChart from "@/components/BarcodeChart.vue";
+import { constructID, SelectionId } from "@/util/OccurenceHighlight";
+import * as d3 from "d3";
+// import * as d3 from "d3";
```
rien, line 90:
```suggestion
```

web/src/components/CompareSide.vue
```diff
@@ -0,0 +1,159 @@
+<template>
+  <div>
+    <component :is="'style'" type="text/css">
+      .marked-code, .token.marked-code {
+        background: var(--markedbg);
+        text-shadow: none;
+      }
+      <template v-for="item in hoveringSelections">
+        .token.operator.{{ item }}, .token.entity.{{ item }}, .token.url.{{ item }},
+        .language-css .token.string.{{ item }}, .style .token.string.{{ item }}, .token.marked-code.{{ item }} {
```
rien, line 10: Why not just use `.{{ item }}` here? Which classes you you want to avoid?

web/src/components/DiffsTable.vue
```diff
@@ -36,6 +36,7 @@ export default class DiffsTable extends Vue {
   @Prop() loaded!: boolean;
   @Prop() diffs!: Diff[];
   @Prop({ default: "" }) search!: string;
+  itemsLoaded = false;
```
rien, line 39: I still don't really get why this is needed. There should only be one point upon which everything is loaded?

web/src/store/index.ts
```diff
@@ -1,28 +1,44 @@
 import Vue from "vue";
 import Vuex from "vuex";
-import { ApiData, fetchData, Diff, Kmer, Metadata, File, ObjMap } from "@/api/api";
+import { ApiData, fetchData, Diff, Kmer, Metadata, File, ObjMap, Block, fetchBlocks } from "@/api/api";
 
 Vue.use(Vuex);
 
+// noinspection SillyAssignmentJS
```
rien, line 7: Is this still needed? I'm not so fond of editor-specific comments in the codebase.

web/src/store/index.ts
```diff
@@ -1,28 +1,44 @@
...
       files: Object() as ObjMap<File>,
       diffs: Object() as ObjMap<Diff>,
-      metadata: Object() as Metadata
-    }
+      metadata: Object() as Metadata,
+    },
+    blocks: Object() as ObjMap<Array<Block>>
+  },
+  getters: {
+    isBlocksLoaded: state => (diffId: number) => {
```
rien, line 20:
```suggestion
    areBlocksLoaded: state => (diffId: number) => {
```

web/src/components/BarcodeChart.vue
ISteampowered: The barcode chart actual display height and width are currently controlled by css. The numbers displayed here are the width and the height of the actual svg coordinate system. Controlling these two number via css is not possible nor does it make much sense.

web/src/components/CompareSide.vue
ISteampowered: prism add a special background to punctuation and characters like `||`. To overwrite this a more specific css selector is needed. These long selectors can be avoided by using `!important` but that is considered bad style.

web/src/store/index.ts
ISteampowered: This was added by accident. I'll remove this

web/src/components/BarcodeChart.vue
rien: If you can choose these values, would it then make sense to use the amount of lines as height and a width of 1?

ISteampowered: Yes it would, this has now been changed

web/src/components/DiffsTable.vue
ISteampowered: I double checked and it is indeed not needed.

rien (approved):

PR #264, opened 2020-08-17
Change output format

### LINKED ISSUE #261: Change output format from dolos analysis
The current output format puts all the fragment data into the `intersections.csv` file, while this is fine for smaller files this can get quite big for more realistic examples. Putting all the fragment data into the csv file has a considerable effect on the web UI as it is forced to load in all the fragments in all at once. Putting the this data into separate files would help resolve these issues.

ISteampowered:
closes #261 
depends on #255

### DIFF
--- a/cli/src/lib/presenter/webPresenter.ts
+++ b/cli/src/lib/presenter/webPresenter.ts
@@ -27,7 +27,6 @@ export class WebPresenter extends CsvPresenter {
       res.contentType("text/json");
       res.setHeader("Content-Disposition", "attachment;filename=blocks.csv");
       this.writeBlocks(res, +req.params.id);
-      // write(res);
     }));
 
     app.use(express.static(path.join(__dirname, "..", "..", "..", "..", "web", "dist")));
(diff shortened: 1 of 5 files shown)


web/src/api/api.ts
```diff
@@ -170,6 +179,11 @@ function parseMetadata(data: d3.DSVRowArray): Metadata {
   );
 }
 
+export async function fetchBlocks(diffId: number, kmers: ObjMap<Kmer>): Promise<Array<Block>> {
```
rien, line 182: See my comment in another PR. I would:
- Keep the `blocks` field in `Diff`, but make it nullable.
- Use this method to populate a `Diff` with its blocks.

cli/src/lib/presenter/webPresenter.ts
```diff
@@ -23,6 +23,11 @@ export class WebPresenter extends CsvPresenter {
     app.get("/data/files.csv", respondwithCSV(o => this.writeFiles(o)));
     app.get("/data/diffs.csv", respondwithCSV(o => this.writeDiffs(o)));
     app.get("/data/kmers.csv", respondwithCSV(o => this.writeKmers(o)));
+    app.get("/data/blocks/:id.json", ((req, res) => {
+      res.contentType("text/json");
```
rien, line 27: 404 handling.

rien (changes requested):
I've mistakenly put comments for this PR somewhere else. In short:
- Keep `blocks` in `Diff` interface and populate them with `populateBlocks`.
- Use a `Map` in `WebPresenter` to fetch diffs by id. Handle 404.

ISteampowered:
These changes are included in #262

PR #265, opened 2020-08-17
Removed unneeded cli options

### LINKED ISSUE #263: Remove unused cli options and rename options to be more logical
Various options are no longer supported in dolos but are still present in the cli. For example the `--directory` and `--base` options

ISteampowered:
closes #263

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -31,22 +31,6 @@ program
     Options.defaultLanguage
   )
   .option(
-    "-b, --base <base>",
-    Utils.indent(
-      "Specifies a base file. Matches with code from this file will never be " +
-      "reported in the output. A typical base file is the supplied code for " +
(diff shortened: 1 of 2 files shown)


cli/src/cli.ts
```diff
@@ -61,27 +45,16 @@ program
...
+      "probably legitimate hash and not the result of plagiarism. With -M " +
       "N any hashing appearing in more than N percent of the files is filtered " +
       "out. Must be a value between 0 and 1. This option is ignored when " +
       "comparing only two files, because each match appear in 100% of the " +
       "files",
     ),
     x => parseFloat(x),
   )
-  .option(
-    "-c --compare",
```
rien, line 75: This option should still be in use?

ISteampowered: It is and has now been re-added.

rien (approved):

PR #274, opened 2020-08-21
Feature/block list

ISteampowered:


ISteampowered:
TODO status:
 - [ ] button spacing in block navigations
 - [ ] change checkboxes to open/closed eye
 - [ ] arrows insead of "next" and "previous"
 - [ ] deselect block when hiding it, or perhaps change highlighting of a selected block that is hidden? 
 - [ ] instead of block start and endline: the amount of matched k-mers and the id?
 - [ ] make sure that the block navigation component does not resize whenever a block is selected/unselected
 - [ ] experiment with placing the block navigation component in a bottom navigation
 - [ ] Remove "hide block" text from toggle panel button in the block navigation component and add a cross to the panel itself that can be used to close it

web/src/components/BarcodeChart.vue
```diff
@@ -1,9 +1,11 @@
 <template >
   <div :id="identifier" class="barcodeChart">
     <component :is="'style'" type="text/css">
-      .marked {
-      fill: var(--markedbg);
-      }
+      <template v-for="item in activeSelections">
```
rien, line 4: Instead of adding a rule for every active selection, is it an idea to add a `.hidden` class to the unactive selections?

web/src/components/BarcodeChart.vue
```diff
@@ -161,13 +164,17 @@ export default class BarcodeChart extends Vue {
 </script>
 
 <style lang="scss">
-@use 'codeHighlightsColours';
+@use 'scssVariables';
```
rien, line 167:
```suggestion
@use 'variables';
```
The fact we're using scss is implied.

web/src/components/BlockList.vue
```diff
@@ -0,0 +1,293 @@
...
+          <v-row>
+            <v-container class="no-y-padding" fluid>
+              <v-row>
+                <v-col class="no-y-padding" cols="12">
+                  <v-list-item-subtitle>
+                    Minimum block length
+                  </v-list-item-subtitle>
+                  <v-slider
+                    :max="highestBlockLength + 1"
+                    :min="lowestBlockLength"
```
rien, line 64: I would rename this to `maxBlockKmers` and `minBlockKmers`.

web/src/components/BlockList.vue
```diff
@@ -0,0 +1,293 @@
...
+</template>
+
+<script lang="ts">
+import BlockVisualizer from "@/components/BlockVisualizer.vue";
+import { Block, Diff } from "@/api/api";
+import { Component, Prop, PropSync, Vue, Watch } from "vue-property-decorator";
+import { constructID, SelectionId } from "@/util/OccurenceHighlight";
+import { SideID } from "@/components/CompareCard.vue";
+
+interface BlockWithId extends Block {
```
rien, line 118: We could add an id to every block?

rien: I thnk blocks already have an id in the CLI, maybe this should be propagated to the frontend as well?

web/src/components/BlockList.vue
```diff
@@ -0,0 +1,293 @@
...
+    if (this.selectedBlock && !this.selectedBlock.active) {
+      this.selectedItem = -1;
+    }
+  }
+
+  changeSelectedItem(dx: number, current?: number): void {
+    if (!this.anyActive) {
+      this.selectedItem = -1;
+    } else {
+      const length = this.diff.blocks!.length;
```
rien, line 194: I would add an explicit assertion to check if something is not undefined. This will also tell typescript it can't be undefined after this.

web/src/components/BlockVisualizer.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <v-list-item
+    class="no-y-padding block-visualizer"
+    dense
+    :disabled="!block.active"
+    :style="dummy ? 'visibility: hidden' : ''"
```
rien, line 6: What is the dummy used for?

web/src/components/BlockVisualizer.vue
```diff
@@ -0,0 +1,71 @@
...
+
+<script lang="ts">
+import { Component, Prop, Vue, Watch } from "vue-property-decorator";
+import { constructID } from "@/util/OccurenceHighlight";
+import { Block } from "@/api/api";
+
+@Component({
+  methods: { constructID }
+})
+export default class BlockVisualizer extends Vue {
```
rien, line 39: Maybe a better name would be `BlockItem`?

web/src/views/Compare.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <v-container fluid fill-height>
     <v-row justify="center">
-      <v-col cols="12">
+      <v-col cols="12" class="no-y-padding">
         <CompareCard
-          v-if="diff"
+          v-if="diff && diff.blocks"
```
rien, line 6:
```suggestion
          v-if="diff?.blocks"
```
If typescript works here.

rien (changes requested):

web/src/components/BlockList.vue
ISteampowered: That is possible however in my code the id is assumed to be the index of the block in the list present within the diff. Maybe I should rename `id` to `index` to avoid confusion?

web/src/components/BlockVisualizer.vue
ISteampowered: `dummy` is used when the component is set to invisible but still needs to take up the same amount of space.

web/src/views/Compare.vue
ISteampowered: I'm afraid the template loader cannot handle typescript no.

web/src/components/BarcodeChart.vue
ISteampowered: This is definitely possible, however then the styling would no longer be completely controlled by the generated css. The other issue is that if a `.hidden` tag would be added programmatically then we would have to check for each span in the document if it has any active selections left before adding the tag, and this would more computationally then letting the css hanlde it.

rien: We'll leave it like this then.

rien (approved):

PR #287, opened 2020-09-01
Hash whole tokens

### LINKED ISSUE #276: Hash whole tokens instead of individual characters of the AST
We now take substrings (kmers) of the stringified abstract syntax tree. Since the length of tokens are very specific to each parser, this makes our default configuration less portable between different languages.

Solution: hash whole tokens at once. 

We use a rolling hash with a fixed amount of characters in it. Since tokens have a variable length we need to add some changes to the hashing. I propose taking the hashes of the whole tokens (possibly memoizing this hash) and using that hash number as input for the rolling hash.

This change will also require some changes in how we keep track of the matching AST Region.

ISteampowered:
closes #276

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -35,12 +35,14 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
+    "@types/sha1": "^1.1.2",
     "chalk": "^4.0.0",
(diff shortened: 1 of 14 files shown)


rien:
Closed by #487.

PR #299, opened 2020-10-06
plagiarism graph

toonijn:
This is a WIP. But I think some feedback and suggestions are welcome.

Selection_002

- If there is an obvious way to categorize students (like program or course), then it would be useful to see this reflected in the colors of the nodes of the graph. How can this categorization be added to Dolos?
- There should be a popup for each node with the student's name, and maybe a link as well (to where?)
- Maybe I should add popups for the edges too. With a small view on the diff of the code of corresponding students. This should link to the compare-view.
- The different components of the graphs are detected. Should this be used?
- Making this graph directed by timestamp of submission is interesting. How can this be implemented in Dolos?


pdawyndt:
Edge weight = similarity? Do edge lengths also have a meaning?

toonijn:
I have made the graph directed and added the possibility to add an info-file (exported from dodona). This info-file should be integrated with the CLI as well.

dolos-web - Google Chrome_007

rien:
Looking great!

Suggestions for improvements:
- Adjust zoom level automatically to fit the nodes. If I load just a few users, it starts too zoomed out.
- Show names next to nodes on hover.
- List 'top collaborators' (with similarity) on the information page.

This will also need a way to navigate to the `/graph/` page, but I'll look into adding a kind of navigation to top bar.

PR #486, opened 2021-04-15
feature/ipynb: Add a ipynb-tokenizer

### LINKED ISSUE #455: Feature Request: Jupyter notebooks
Jupyter notebooks (.ipynb) are a combination of text and code (in possibly different languages). It would be nice if Dolos could be used to find plagiarism between such notebooks, in text as well as in code.

toonijn:
This is a first prototype of a ipynb-tokenizer (and markdown-tokenizer). Before finalizing this, I have a few question:
- Can I customize the diff-view in the web-format?  The problem is visible in the following image. Can I specify a different text-file to be displayed?
image
- The function newTokenizer https://github.com/dodona-edu/dolos/blob/09f7a252fde5e7013b263b9c4ae001a64118ad0e/cli/src/dolos.ts#L17 is getting awkward. Suggestions?

This fixes #455

### DIFF
--- a/cli/src/dolos.ts
+++ b/cli/src/dolos.ts
@@ -10,11 +10,17 @@ import * as path from "path";
 import { Tokenizer } from "./lib/tokenizer/tokenizer";
 import { CharTokenizer } from "./lib/tokenizer/charTokenizer";
 import { default as fsWithCallbacks } from "fs";
+import { IPyNbTokenizer } from "./lib/tokenizer/ipynbTokenizer";
+import { MarkdownTokenizer } from "./lib/tokenizer/markdownTokenizer";
 const fs = fsWithCallbacks.promises;
 
 function newTokenizer(language: string): Tokenizer {
   if (language == "chars") {
     return new CharTokenizer();
+  } else if (language == "ipynb") {
+    return new IPyNbTokenizer();
+  } else if (language == "md") {
+    return new MarkdownTokenizer();
   } else if (CodeTokenizer.supportedLanguages.includes(language)) {
     return new CodeTokenizer(language);
   }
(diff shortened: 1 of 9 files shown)


cli/src/dolos.ts
```diff
@@ -10,11 +10,17 @@ import * as path from "path";
...
+import { IPyNbTokenizer } from "./lib/tokenizer/ipynbTokenizer";
+import { MarkdownTokenizer } from "./lib/tokenizer/markdownTokenizer";
 const fs = fsWithCallbacks.promises;
 
 function newTokenizer(language: string): Tokenizer {
   if (language == "chars") {
     return new CharTokenizer();
+  } else if (language == "ipynb") {
+    return new IPyNbTokenizer();
+  } else if (language == "md") {
```
rien, line 22: The alternative here would be to make something like
```typescript
const tokenizers = new Map();
tokenizers.add("chars", () => new CharTokenizer());
tokenizers.add("ipynb", () => new IPyNbTokenizer());
...

const tokenizer = tokenizers.get(language);
if (tokenizer) {
  return tokenizer.call();
} else {
  return new CodeTokenizer(language);
}
```
But this does not really remove complexity. A switch would also make sense here because of the `default`.

rien (commented):
If you want the displayed content to be different than the actual content, you can change the `IPyNbTokenizer`'s `tokenizeFile` to output a subclass of `TokenizedFile`which overwrites the `content()` getter with what you want it to look like. This function is used in the `CsvPresenter` to pass the file contents to the web view.

I would maybe not overwrite this getter, but add a `public display(): string` method to `File` which by default just returns the normal content, but can be overwritten by a `IPyNbTokenizedFile`.

To improve readability, you could also swap the occurence of `IPyNB` to `Notebook` (e.g. `NotebookTokenizer`).

toonijn:
After the rebase on develop, it does work:
There was no plagiarism in my course!

rien:
As this PR is wildly outdated and probably won't receive any updates soon, I will close it. I've added an issue referring to this prototype, so we don't lose the progress made here.

PR #487, opened 2021-04-16
Hash AST tokens instead of slicing serialized AST

### LINKED ISSUE #276: Hash whole tokens instead of individual characters of the AST

rien:
This is a major rework in the internals: instead of slicing kmers out of the serialized AST tree, we instead hash the individual tokens present in the AST. A kmer is now a series of k subsequent tokens. This is much more logical because it respects the token boundaries.

This is a rework of #287 which uses more efficient hashing (the same rolling hash, except it is not rolling) and simplifies how our filters reconstruct debug data.

Closes #276 and #287


cli/src/lib/hashing/tokenHash.ts
```diff
@@ -0,0 +1,17 @@
...
+    this.mod = 33554467;
+    this.base = 5801;
+  }
+
+  public hashToken(token: string): number {
+    let hash = 0;
+    for (let i = 0; i < token.length; i++) {
+      hash = (this.base * hash + token.charCodeAt(i)) % this.mod;
+    }
+    return hash;
```
rien, line 15: @toonijn can you check if this makes sense as a hashing algorithm? This is essentially the same hashing algorithm as the one used in [rollingHash.ts](https://github.com/dodona-edu/dolos/blob/hash-tokens/cli/src/lib/hashing/rollingHash.ts#L35-L42), except it is not rolling.

This will be used to hash tokens, which will usuallyl be very short (< 20 chars) strings.

toonijn:
Maybe we need to find a more permanent place for this comment (say an article).
Some thoughts about hashing and our hash function:

- Many articles and papers provide strong cryptographically secure hash functions. These papers our of little interest to our goals. We need fast with little collisions. We do not care if an adversary could carefully construct a hash. Worst case: his code will be flagged as plagiarism.
- Many articles, especially older ones, are great advocates for shift and xor types of hash functions. These are very efficient, but the chosen constants are harder to get right. Especially due to the fact the javascript numbers are a little funny.
- Avoiding collisions is the goal. Javascript has some funny quirks, which we need to work around:
    - Numbers are always double precision, so 53 bits of precision.
    - It would be ideal if our hash would also contain 53 bit. But in practice, one likes to multiply two hashes to construct a new, commutative hash. For example when hashing a set, are constructing an incremental hash function.
    -  To enable us to take the product of two hashes we would need numbers with 2*T bits (with T the number of bits in our hash). As the largest numbers in Javascript are 53 bit. The largest hash we could take would be 26 bit (less then 70 million different values).
- We need a rolling hash functions. This limits the possibilities immensly. 

When taking a deep dive into the literature one comes across [A Design Principle for Hash Functions](https://link.springer.com/content/pdf/10.1007%2F0-387-34805-0_39.pdf). This paper mainly consists of cryptographically secure functions. Even quite simple hash functions from this paper (e.g. section 4.1), are not suited for rolling hashes.

Another paper, which primarily focuses on string hashing, is [Performance in practice of string hashing functions](http://www.quretec.com/u/vilo/edu/2003-04/DM_seminar_2003_II/ver1/P12/articles/ramakrishna97performance.pdf). This article brings forth a very useful way of looking at the problem:
image
for certain `init`, `step`, and `final` functions. Because we want a rolling hash function, the step function should be quite manageable, yet still result in a useful hash. They give some examples, the one we have chosen for can be found as the first example in section 5. One of the big reason we do like this method is its ability to be used as a rolling hash. Let c_i be the hash of the i'th token, the hash h_(i,k) of length k from i onward is defined as

h_(i,k) =  (c_i r^(k-1) + c_(i+1) r^(k-1) + ... + c_(i+k-2) r + c_(i+k-1)) % p

In this formula is r a well chosen radix and p a well chosen large prime. Now the rolling property is as

h_(i+1, k) = (c_(i+k) + r h_(i,k) - r^k c_i ) % p

The value r^k can be precomputed, which makes this formula efficiently computable.

So if p has T bits, our hash will have as well. r should have as many bits R as possible, such that already with few character, the full output range of bits is used, independent of how many bits c_i has.  But, ignoring the rolling property for now, on the other hand, to fully maximize the 53 bit range, it is sufficient if R+T < 53. Then everything will still be computable with Javascript's numbers. Ideally This would yield, for example, T = 34 and R = 17. But if the rolling hash feature is required r^k probably will have the full T bits.  (It is possible to find a r such r^k would have a lot fewer bits. This makes r dependent on k, which is in practice very cumbersome to implement.)

Lastly, very concretely: picking p and r. As p² should be representable in Javascript, any prime number with at most 26 bit will suffice. As for choosing r, the order of this number modulo p should be maximal, which always is the case if p is prime. r should also not be a power of 2 and as large as possible. 



@rien As an answer to your question: I believe the base value r should be something larger (in your code as well as in the rolling hash). Just pick any number with 20-something bits, and you are golden. Maybe, I have a feeling, for the modulus p a random prime between 2^25 and 2^26 would be better than the one it is now. 
https://github.com/dodona-edu/dolos/blob/edc033c20a24ade757b4ce6bb61157b5bf375cd3/cli/src/lib/hashing/rollingHash.ts#L25

toonijn:
Maybe a small note:
the hash for a token, and the rolling hash should have different bases r.
Else ['abc'], ['ab', 'c'], ['a', 'bc'], ['a', 'b', 'c'] would result in the same hash.

cli/src/lib/hashing/tokenHash.ts
```diff
@@ -0,0 +1,43 @@
...
+   * as possible.
+   *
+   * Hence this is the largest prime number such that 127 * base < mod.
+   */
+  readonly base: number = 747287;
+
+  public hashToken(token: string): number {
+    let hash = 0;
+    for (let i = 0; i < token.length; i++) {
+      hash = (this.base * hash + token.charCodeAt(i)) % this.mod;
```
toonijn, line 39: The hash of a single token is simply it's ascii value. This should be fixed.

rien: Yes, I was thinking this is not good. Any suggestions how to fix it?

toonijn:
```javascript
      hash = (hash + token.charCodeAt(i)) * this.base % this.mod;
```

toonijn: This is may also be a problem for the rolling hash function.

rien: It is, but since the rolling hash will be used with the output hashes of the token hash, we should be using all the bits (with the fix you proposed).

rien: Another fix could be to add an initial (or final) step to multiply by a seed number which uses a lot of bits.

PR #599, opened 2021-08-19
Deploy demo page after each new release

rien:
This changes the release workflow such that after each release, the demo page is updated on our website (https://dolos.ugent.be/try/).


.github/workflows/release.yml
```diff
@@ -88,3 +88,63 @@ jobs:
...
+    steps:
+      - name: Setup Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: "16"
+      - name: Install dolos
+        run: |
+          npm install -g @dodona/dolos
+          dolos --version
+      - name: Setup SSH
```
bmesuere, line 103: Any reason why you do this manually instead of using `shimataro/ssh-key-action@v2` ?

bmesuere (approved):

.github/workflows/release.yml
rien: Simplicity, mostly. I've looked at that action but it is essentially equal to the five lines in this step. Adding an extra dependency on an action for such a simple task  seemed like overkill to me.

PR #622, opened 2021-09-02
Add a --port specifier to the CLI

toonijn:
This PR allows to specify a port on which the webserver should listen.
```
yarn run dolos --format=web --port=3001
```

This is useful when port 3000 is already used, or when multiple exercises are analysed simultaneously.


rien:
This change is now included in #621, I have added you as an author.

PR #660, opened 2021-10-10
Feature/641 clustering

### LINKED ISSUE #641: List of clusters
We should have clear overview of the clusters of plagiarized submissions.

This will require a clustering algorithm (possibly based on the threshold #640) and a clear way to show the results.

AurisAudentis:
This is a pull request to add a list of clusters (fixes #641). This PR focuses mainly on the visualization/interaction part of things by adding a list component and adding graph color view.

It contains only one clustering algorithm (Single Linkage) for now, but it is written to be extensible (common, utility functions lifted to ClusterFunctions.ts).  I propose that further algorithms are added as future PR's.

## Statistics
Currently there's an average similarity metric and a size metric, are there other metrics that might be considered useful? Please leave suggestions.

## Todo
- [x] Add visualisation of clusters


pdawyndt:
Is this "single linkage": a document is added to a cluster as soon as it has a similarity above the threshold with at least one other document in the cluster?

AurisAudentis:
> Is this "single linkage": a document is added to a cluster as soon as it has a similarity above the threshold with at least one other document in the cluster?

For now, yes. I'm planning to rework this feature a little bit to make multiple clustering algorithms easy to implement, but the focus of this PR is mostly the visualization by coloring the graph view and adding a list component. I was planning to propose a different algorithm (based on common similarity) on Thursday to implement in a different PR.

AurisAudentis:
As discussed on thursday, I updated the list to a different view with collapsibles. It still requires some work to make it visually appealing, but I'd like some feedback before I get started on that: is this the direction we want to go?

AurisAudentis:
Currently, I have two versions of the timeline: one with a selection tool, and one with a classic popup. If possible, I would like some feedback on which version to use:
image
image

@rien  : could we deploy the one with the selection tool? The popup tool is probably more intuitive.
@bmesuere: do you have a minute to give feedback on the timeline, and the clusters in general? I'm trying to wrap this PR up. Thanks!

rien:
Some remarks I mentioned in previous meeting:
- Maybe this visualization for the clustering view is not ideal, because it is yet another table/list. It does not really give a good "first impression": you have to expand a card to view its contents.
- An alternative idea I mentioned: show the clustering with the plagiarism graph (e.g. group the clusters together in an orderly way, and show information about that cluster).
-I think this branch (with the selection tool) is actually best in practice. I suggested the tooltip earlier, but it is much easier to just drag along the timeline.
- I still think the _graph view_ and _pair view_ links are too similar to the three tab-buttons to the right. Maybe a different location or just different styling could resolve this.

PR #661, opened 2021-10-10
Web dynamic fragment loading

ArneCJacobs:
Depends on #659 
This PR enables the `web` package to calculate the needed fragments whenever needed.
Concrete changes
- `web` package
    - the `library` package is imported 
    - fragment data is no longer read from files, but instead calculated by calling the `library` package using already tokenized files so that `tree-sitter` is not used
    - webpack is explicitly configured so that `tree-sitter` and `codeTokenizer` are always excluded
- `cli` package
    - the default behaviour of `FileView.writeToDirectory` now no longer writes fragments unless the `writeFragments` argument is set to true
    - `FileView.show()` still writes fragments to ensure compatibility with 3rd party tools
    - `FileView.writeFiles` now also writes mapping and the ast is now stored as an array instead of a concatenated string
    - `FileView.writeMetadata` now also writes types as csv does not preserve them
 - `library` package
     - `tree-sitter` is now an optional dependency
     - `CodeTokenizer` is now loaded dynamically 
     - `Dolos.analyzeTokenized` now allows you to analyze tokenized files directly 
     
since the `web` package now depends on `library`, that needs to be first. However, dispite thing being a typescript project, project references cannot be used without breaking `vue-cli-service build` and thus build order cannot be directly enforced.


ArneCJacobs:
@rien This pull is ready now.

rien:
Can you rebase this PR with the main branch? There are some changes that are not needed anymore (GitHub actions workflow, lib package entrypoint, ...).

web/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
-    "lint": "vue-cli-service lint"
+    "lint": "vue-cli-service lint",
```
AurisAudentis, line 9: If we are adding a linter, we should probably also incorporate it in the CI.

web/src/api/api.ts
```diff
@@ -231,7 +231,12 @@ export async function loadFragments(pair: Pair, kmers: ObjMap<Kgram>, customOpti
     [fileToTokenizedFile(pair.leftFile), fileToTokenizedFile(pair.rightFile)]
   );
   const reportPair = report.scoredPairs[0].pair;
-  pair.fragments = parseFragments(reportPair.fragments(), kmers);
+  const hashToKmers: ObjMap<Kgram> = {};
```
AurisAudentis, line 234: I generally prefer using [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), but maybe we should have a unified style for this?

web/public/data/kgrams.csv
```diff
@@ -1,57 +1,86 @@
...
-58,3995583,,"[7,6]"
-59,2683139,,"[7,6]"
-60,581367,,"[7,6]"
-61,1077799,,"[7,6]"
-62,3366137,,"[7,6]"
-63,165354,,"[7,6]"
+8,1083181,,"[6,4,7]"
+9,2676335,,"[6,4,7]"
+10,6644131,,"[6,4,7]"
+11,604601,,"[6,4,7]"
```
AurisAudentis, line 5: Do we want to change the public/data files?

web/src/api/api.ts
```diff
@@ -192,15 +199,44 @@ function parseKgrams(kgramData: d3.DSVRowArray, fileMap: ObjMap<File>): ObjMap<K
   }));
 }
 
+function castToType(rowValue: string | null, rowType: string): string | boolean | number | null {
```
AurisAudentis, line 202: You can make this function better typed by using type discrimination.

AurisAudentis (commented):
I had a quick look at the code and have a couple of suggestions.

web/package.json
ArneCJacobs: The linter was present and is also already used in the ci. This change added a comma so the "serve:watch" command could be added.

web/public/data/kgrams.csv
ArneCJacobs: These changes are needed for dynamic fragment loading. All changes have been discussed an approved by @rien

web/package.json
AurisAudentis: Oops, my bad :facepalm:

lib/package.json
```diff
@@ -37,6 +37,9 @@
     "typescript": "^4.3.2"
   },
   "dependencies": {
+  },
+  "optionalDependencies": {
+    "tree-sitter": "^0.19.0",
```
rien, line 42: It seems like two versions of `tree-sitter` are added now?

web/src/api/api.ts
```diff
@@ -114,35 +125,31 @@ async function fetchMetadata(
...
-}
-
 function parseFiles(fileData: d3.DSVRowArray): ObjMap<File> {
   return Object.fromEntries(
     fileData.map(row => {
       const extra = JSON.parse(row.extra || "{}");
       extra.timestamp = extra.createdAt && new Date(extra.createdAt);
       row.extra = extra;
+      row.mapping = JSON.parse(row.mapping || "[]");
+      row.ast = JSON.parse(row.ast || "[]");
```
rien, line 135: I'm not 100% sure, but this might be cause the UI to hang for a while when loading a lot of files (parsing JSON is notoriously slow). Can you check if this is the case?

If this is indeed causing the UI to hang, we might consider delaying the `JSON.parse` to when a file is actually needed.

rien (commented):
I have tested this locally, and this is working great. The analysis itself finishes a lot quicker now and the pairwise view is now shown almost instantly.

- [x] As you mentioned, there is still a bug where the first pairwise view you open is not rendered correctly.
- [ ] When opening an analysis with a lot of files, Dolos hang for a while (10s for SOCO Java). I've launched the profiler and it seems it is doing a lot of allocations (and GC's).  I've added a comment where I _think_ this is caused, but you should check this to make sure.

ArneCJacobs (commented):
~~Don't merge quite yet, there is a bug causing some variables to be uninitialized in CompareSide.~~

This issue is not resolved

rien (approved):

PR #686, opened 2021-12-22
feat(host): added new option to define a host

radim-kliment:
This pull request adding a new option to define a host. Previously default host is localhost which is problematic when you want work with docker container and don't want to use a host network.

### DIFF
--- a/cli/src/cli/commands/run.ts
+++ b/cli/src/cli/commands/run.ts
@@ -99,6 +99,14 @@ export function runCommand(program: Command): Command {
       "3000"
     )
     .option(
+      "-H, --host <host>",
+      Utils.indent(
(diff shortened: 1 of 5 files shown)


rien:
Hi @radim-kliment, thank you for your PR, this will indeed allow one to omit the host network on docker.

How do you plan to pass the host-parameter? Will you assign an ip (or hostname) to the docker container itself, or will you set `-H 0.0.0.0` and visit the assigned ip of the docker container itself?

radim-kliment:
For now I'am exposing a port of a container and pass a 0.0.0.0 to the script by this parameter. After that Iam using a Traefik reverse proxy to handle redirects from defined subdomain to the container.

rien (approved):

PR #691, opened 2022-01-24
Add explanation of the analysis parameters

rien:
This PR adds a section to the documentation to explain the function and impact of the `-k` and `-w` parameters.

The other parameters are intentionally omitted because these parameters can change in the future. Instead, I refer to the output of `dolos --help`.

If we want a complete overview of the command-line options on the documentation site, this could be generated using the code in [run.ts](https://github.com/dodona-edu/dolos/blob/main/cli/src/cli/commands/run.ts) and [options.ts](https://github.com/dodona-edu/dolos/blob/main/lib/src/lib/util/options.ts). But this will be a lot of work for something that is already available.

### DIFF
--- a/docs/guide/running.md
+++ b/docs/guide/running.md
@@ -42,11 +42,11 @@ The analysis parameters can be altered by passing the relevant program
 arguments. For more information about the impact of these parameters you can
 read the section describing [how Dolos works](./algorithm.html).
 
-If you are having performance issues with Dolos, it can help to alter these
-parameters. It is recommended to increase the Window length (`-w`) first before
-altering the _k_-gram length (`-k`).
+If a Dolos analysis is taking too long or consuming too much memory, it can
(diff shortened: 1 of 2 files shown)


docs/guide/running.md
```diff
@@ -36,6 +36,45 @@ You can show all the command-line options by passing the `-h` or `--help` flag
 or by running `dolos help run`.
 :::
 
+## Modifying analysis parameters
+
+The analysis parameters can be altered by passing the relevant program
+arguments. For more information about the impact of these parameters you can
+read the section describing [how Dolos works](./algorithm.html).
+
+If you are having performance issues with Dolos, it can help to alter these
```
bmesuere, line 45: Maybe specify which performance you are talking about: accuracy or runtime

bmesuere (approved):

PR #717, opened 2022-03-10
Feature/similarity cutoff interpolation

AurisAudentis:
This feature adds an algorithm to find a good default value for the similarity cutoff, by using different observations about the likely value.

We pick a local minimum around the 0.8-value, which often results in a decent starting value. This value can still be modified by the slider.


web/src/store/api.ts
```diff
@@ -103,3 +105,59 @@ export default {
     }
   }
 };
+
+function getInterpolatedSimilarity(pairs: Pair[], step = 0.03): number {
+  pairs.sort((p1, p2) => p1.similarity - p2.similarity);
+  const binnedCount = getBinnedCount(pairs, step);
+  console.log(binnedCount);
```
rien, line 112: There are some `console.log` statements that should probably be removed.

rien (approved):
Can you add some documentation with a short description of what the algorithm begind the interpolation is? Otherwise this is :+1:

PR #718, opened 2022-03-10
Feature/summary

AurisAudentis:
This feature adds


rien:
Some remarks mentioned previous meeting:

Header: 
- Let the "facts" from the header (highest similarity, longest match) link to a relevant page.
- The three buttons are currently used as a "filler" and I think something more actionable may be relevant (for example: compare the highest pair(s), show a clustering, ...).
- The graphs could use a "line" at the location of the mouse that show the current value of the bar below

Cards:
- There are now different cards for different metrics (similarity, longest fragment, ...). We need a visual representation to make this clear. For example: each card could have tabs which are immediately visible (and clicking on these tabs would also show the other info).
- Navigation to the next batch of cards is a bit clunky, a (large) button could be added to the bottom "Next cards".
- If there is no metadata available, it displays as "unknown". I would hide this instead if we cannot show something useful.
- Each kind of card essentially shows the same graph, which is not ideal. Maybe this information can be shown on a "central" graph (which is the same for each card), but which highlights where that file is situated on that graph.

PR #726, opened 2022-04-12
Semantic analysis of matches

AurisAudentis:


web/src/components/CompareSide.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div style="height: 100%">
```
rien, line 2: Can this be moved to the `<style>` block in this component?

web/src/components/FragmentList.vue
```diff
@@ -316,9 +330,9 @@ export default class FragmentList extends Vue {
   @Watch("selectedItem")
   onSelectedItemChange(newVal: number): void {
     if (newVal === -1) {
-      this.dataTableSelection = [];
+      // this.dataTableSelection = [];
     } else {
-      this.dataTableSelection = [this.fragmentsWithId[newVal]];
+      // this.dataTableSelection = [this.fragmentsWithId[newVal]];
```
rien, line 335: This function doesn't seem to do anything anymore, should it be removed?

rien (changes requested):
Unfortunately, loading the page is now very slow and is being unresponsive in the process. While this is a good addition to Dolos itself, I don't think it is worth giving up the responsiveness.

Do you think it would be possible, to make this optional or calculate this in the background? If not, I think we should try speeding this up as much as possible, or reconsider moving this calculation to the CLI.

The compare view seems to be broken in some cases (no highlighted parts, and the code blocks are shown fully instead of in a scroll block).

lib/src/lib/analyze/index.ts
```diff
@@ -8,15 +8,16 @@ import { WinnowFilter } from "../hashing/winnowFilter";
...
 import { TokenizedFile } from "../file/tokenizedFile";
+import { SemanticAnalyzer } from "./SemanticAnalyzer";
 
 type Hash = number;
 
 export class Index {
   private readonly kgramLength: number;
   private readonly kgramsInWindow: number;
-  private readonly index: Map<Hash, Array<Occurrence>> = new Map();
+  //private readonly index: Map<Hash, Array<Occurrence>> = new Map();
```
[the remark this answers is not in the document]
AurisAudentis: Yes, using this setup with local variables was necessary to enable the instance of this class to be re-used in SemanticAnalyzer to annotate the AST's as required for the algorithm.

lib/src/lib/analyze/SemanticAnalyzer.ts
```diff
@@ -0,0 +1,292 @@
...
+
+export type UnpairedNodeStats = LeftOnly | RightOnly;
+
+type ReturnData = {
+  nodeStats: NodeStats;
+  currentI: number;
+  lastMatchedLevels: Map<number, NodeStats[]>;
+}
+
+export class SemanticAnalyzer {
```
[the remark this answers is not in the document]
AurisAudentis: This analysis code is not specific to the browser (it is not responsible directly for visualisation) or to the CLI code (running the analysis). It is display-independent algorithmic code, which I felt fits better in the platform-independent 'lib'.

I still remain convinced the 'lib' package is the appropriate place to store this, especially considering the step of moving the execution of the algorithm to the CLI at analysation time. Do you feel it should be moved?

cli/src/cli/commands/run.ts
```diff
@@ -154,6 +154,23 @@ export function runCommand(program: Command): Command {
       x => parseFloat(x),
       Options.defaultKgramsInWindow
     )
+    .option(
+      "--semantic",
+      Utils.indent(
+        "Using the semantic analysis augment in Dolos",
```
rien, line 160:
```suggestion
        "Enable a semantic analysis on the matched fragments",
```

cli/src/cli/commands/run.ts
```diff
@@ -154,6 +154,23 @@ export function runCommand(program: Command): Command {
...
+        Options.defaultSemantic
+      ),
+      Options.defaultSemantic
+    )
+    .option(
+      "--semantic-match-length <number>",
+      Utils.indent(
+        "Length of the smallest semantic match",
```
rien, line 168: Is this the length in tokens?

lib/src/lib/analyze/SemanticAnalyzer.ts
```diff
@@ -0,0 +1,292 @@
...
+    const results = [];
+    const astMap = await this.astWithMatches(tokenizedFiles, hashFilter);
+    for(const tokenizedFile of tokenizedFiles) {
+      results.push(await this.semanticAnalysisOneFile(tokenizedFile, astMap.get(tokenizedFile)));
+    }
+
+    return [astMap.get(tokenizedFiles[0]).groups, results];
+  }
+
+  private async semanticAnalysisOneFile(
```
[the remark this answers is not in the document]
rien: I think this is still something that can be changed?

lib/src/lib/analyze/index.ts
```diff
@@ -84,7 +85,28 @@ export class Index {
...
+  }
+
+  public async createMatches(
+    tokenizedFiles: TokenizedFile[],
+    hashFilter = this.hashFilter
+  ): Promise<Map<Hash, Array<Occurrence>>> {
+    const index = new Map();
+
+    // TODO REMOVE TEST
+    tokenizedFiles.forEach(t => t.kgrams.splice(0, t.kgrams.length));
```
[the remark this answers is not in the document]
rien: I think this doesn't do anything?

web/src/components/CompareSide.vue
```diff
@@ -36,6 +40,7 @@ import "prismjs/themes/prism.css";
 import "prismjs/plugins/line-numbers/prism-line-numbers.js";
 import "prismjs/plugins/line-numbers/prism-line-numbers.css";
 import { ID_START, registerFragmentHighlighting } from "@/util/OccurenceHighlight";
+import { NodeStats } from "@dodona/dolos-lib/dist/lib/analyze/SemanticAnalyzer";
```
rien, line 43:
```suggestion
import { NodeStats } from "@dodona/dolos-lib";
```
And `NodeStats` will probably need to be exported by the lib.

rien (changes requested):
- There are some `yarn.lock` files and `package-lock.json` files added that should be removed
- It seems you changed the files in the `web/public/data` with python files analyzed with the language parser? Is there a specific reason for this?

rien (approved):
:+1:

PR #734, opened 2022-05-04
fix 733: fragments folder check

AurisAudentis:
This PR removes a legacy check for the fragments folder, which is no longer in use since @ArneCJacobs added the dynamic caluclation a few months ago.


rien (approved):
Perfect, thank you for the quick fix!

PR #736, opened 2022-05-12
Update require for now scoped tree-sitter-elm package

klassiker:
The `tree-sitter-elm` was scoped to `@elm-tooling/tree-sitter-elm`, the unscoped package has not been updated for over a year now and uses an older version of `tree-sitter`. See: https://www.npmjs.com/search?q=tree-sitter-elm

I'm not sure about the best way to handle this, maybe a lookup dictionary for known / tested languages? This would obviously require some maintenenace.

Another approach would be to check if `language` starts with `@` or contains `/`, in which case `language` should be interpreted as a package name.

I've included my quickfix here in case you don't want to make a decision on that (yet).

### DIFF
--- a/lib/src/lib/tokenizer/codeTokenizer.ts
+++ b/lib/src/lib/tokenizer/codeTokenizer.ts
@@ -26,7 +26,11 @@ export class CodeTokenizer extends Tokenizer {
    */
   public static registerLanguage(language: string): void {
     try {
-      require("tree-sitter-" + language);
+      if (language === "elm") {
+        require("@elm-tooling/tree-sitter-elm");
+      } else {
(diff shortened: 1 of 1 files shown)


rien (approved):
Hi @klassiker, this is looking good.

Have you tested if Dolos is working properly with Elm? If so, we can add Elm to the list of "officially" supported languages and ship Elm support with the default CLI.

On a side note: if tree-sitter plugins are evolving to using non-standard names, we might consider using a `Map` instead of if-statements if more languages would do this.

rien:
Perfect, I'll create an issue to add Elm to our default supported programming languages then.

Thank you for your contribution!

PR #741, opened 2022-05-28
Pseudonymize dynamically

AurisAudentis:
This PR adds a toggle in the top right (should probably be moved, please feel free to suggest a better spot) which pseudonymizes the dataset for use in presentations and reports. 

Real names and submissions may still be gotten from the code itself (which is not filtered) and identification interpolated from the timestamp, which is only moved by a global pseudorandom offset to maintain the relative time of the entries. This relative time is important for some visualisations, so cannot be fully offset.


web/src/App.vue
```diff
@@ -54,6 +55,13 @@
             <v-list-item-content>File Analysis</v-list-item-content>
           </v-list-item>
         </v-list>
+
+        <template v-slot:append>
+          <div class="pa-2" v-if="!isCollapsed">
```
rien, line 60: This does "hide" the pseudonymization feature a bit.

Maybe we could show an "incognito" symbol in the location of the toggle button if the navbar is collapsed:
image

If the user would hover over this icon, the navbar will expand and the toggle button will become visible.

web/src/App.vue
```diff
@@ -102,3 +112,9 @@ export default class App extends Vue {
   }
 }
 </script>
+
+<style>
+.v-messages {
+  display: none;
```
rien, line 118: Which messages are we hiding?

rien (changes requested):
When the anonymization is toggled, the UI hangs for a moment and the files are fetched over network again.

I think we can improve this by always generating the pseudonymized data when parsing the files, but letting the `isAnonymous` flag decide which data is shown. This will make toggling the anonymization more instant.

web/src/App.vue
AurisAudentis: The slider component provides extra space for some kind of messages to be added. In our case, however, it causes a bit of a style mess.

AurisAudentis:
> When the anonymization is toggled, the UI hangs for a moment and the files are fetched over network again.
> 
> I think we can improve this by always generating the pseudonymized data when parsing the files, but letting the `isAnonymous` flag decide which data is shown. This will make toggling the anonymization more instant.

How do you see this implemented in practice? Do you suggest storing the files in the state twice, once anonymized and once not anonymized, or would you duplicate fields in the files themselves?

I would prefer that anonymization works 'transparently' for the visualisations, that is: they do not need to react to whether the files are anonymized or not. This eliminates some risk of misuse.

rien:
I would store this in the file itself and have the `name` and `time` getters return a different value depending on the current `isAnonymous` setting.

web/src/App.vue
```diff
@@ -81,6 +97,16 @@ export default class App extends Vue {
     }
   }
 
+  setCollapsed(o: boolean):void {
+    console.log(o);
```
rien, line 101: Remove `console.log`.

web/src/App.vue
```diff
@@ -81,6 +97,16 @@ export default class App extends Vue {
...
+    this.isCollapsed = o;
+  }
+
+  @Watch("isCollapsed")
+  print():void {
+    console.log(this.isCollapsed);
```
rien, line 107: Remove `console.log`.

rien (approved):

PR #764, opened 2022-06-22
Refactor graph components

maartenvn:
* Refactor the following components to the Composition API:
  * GraphView
  * ClusteringTable
  * ClusteringCard
  * GraphSelectedInfo
* Introduce `useCluster` composable with cluster helpers.

### DIFF
--- a/web/src/components/ClusteringTable.vue
+++ b/web/src/components/ClusteringTable.vue
@@ -58,7 +58,7 @@ export default defineComponent({
     const route = useRoute();
     const { cutoff } = storeToRefs(useApiStore());
 
-    // Active expension panel.
+    // Active expansion panel.
     const panel = ref(-1);
 
     // Table headers


web/src/components/ClusteringTable.vue
```diff
@@ -2,98 +2,117 @@
...
-  @Prop() loaded!: boolean;
-  @Prop() currentClustering!: Clustering;
-  @Prop({ default: "" }) search!: string;
+  setup(props) {
+    const vuetify = useVuetify();
+    const route = useRoute();
+    const { cutoff } = storeToRefs(useApiStore());
 
-  public panel = -1;
+    // Active expension panel.
```
rien, line 61:
```suggestion
    // Active expansion panel.
```

web/src/components/clustering/ClusteringCard.vue
```diff
@@ -2,96 +2,111 @@
...
+    // If the timeline should be shown.
+    // Timeline will only be shown if every cluster element has a timestamp.
+    const showClusterTimeline = computed(() => {
+      return getClusterElementsArray(props.cluster).every(
+        (e) => e.extra?.timestamp
+      );
+    });
+
+    // String of ids to show in the pair view.
+    // TODO: it may be better to inline the pair-table in the future for better UX.
```
rien, line 88: Can you make an issue instead of this TODO? This will make it easier to track this.

web/src/util/Types.ts
```diff
@@ -1 +1,4 @@
+import { Ref } from "@vue/composition-api";
+
 export type SortingFunction<T> = (a: T, b:T) => number;
+export type MaybeRef<T> = T | Ref<T>;
```
rien, line 4: Your editor seems to be omitting newlines on the end of a file, can you change this?

Maybe we should also add an `.editorconfig` to define our preferences.

PR #766, opened 2022-06-24
Refactor compare view & code viewer

maartenvn:
* Refactor CompareCard to the Composition API
* Refactor CompareCard to the Composition API
* Refactor SemanticList to the Composition API
* Fix bug introduced in previous refactor.

These components do have some minor bugs and performance issues, but these will be fixed in a future rewrite of the code viewer when migrating to a different library.

### DIFF
--- a/web/src/components/CompareSide.vue
+++ b/web/src/components/CompareSide.vue
@@ -1,160 +1,226 @@
 <template>
   <div class="main-container">
     <component :is="'style'" type="text/css">
-
       <template v-for="item in activeSelections">
-        .token.marked-code.{{item}} {
-        background: var(--markedbg);
-        text-shadow: none;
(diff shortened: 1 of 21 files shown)


web/src/components/CompareSide.vue
```diff
@@ -1,160 +1,226 @@
 <template>
   <div class="main-container">
     <component :is="'style'" type="text/css">
-
       <template v-for="item in activeSelections">
-        .token.marked-code.{{item}} {
-        background: var(--markedbg);
-        text-shadow: none;
-        }
+        .token.marked-code.{{ item }} { background: var(--markedbg);
```
rien, line 5: This has some weird formatting, although it is CSS.

PR #779, opened 2022-06-28
Fix anonymous toggle in sidebar

maartenvn:
Fix anonymous toggle in sidebar


rien (changes requested):
You seem to have accidentally committed `web/yarn.lock`, can you remove this file?

PR #783, opened 2022-06-29
Improve Github Actions & linter configuration

maartenvn:
Improve Github Actions & linter configuration.

- [x] Fix multiple linter warnings due to the different Node versions
- [x] Fix web linter not outputting in the correct format
- [x] Split Github actions CI workflow into multiple scoped workflows
- [x] Only trigger scoped workflow when changes are detected


maartenvn:
@rien This is ready 🙂

rien (changes requested):
- [x] ~~I don't see changes to the release workflow, but it also uses the same dependencies setup. Can you change this as well? Be careful not to publish a new version though.~~ I will take a look at this when making a new release.

- [x] ~~You can use [yaml aliases](https://yaml101.com/anchors-and-aliases/) to avoid some duplication.~~ It is not possible to merge an alias into a yaml list, so we cannot use this.

- [x] Can you check if this would [stop branches from being mergeable](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks) because some checks are not completed?


Since it is now split into multiple workflows, there is less potential to use aliases and anchors, and we might have trouble with the comment I made above. Can you explore if it would maybe still be possible to use a conditional check on a job instead of a workflow?

rien (approved):

PR #784, opened 2022-06-30
Refactor to the Composition API

maartenvn:
* Refactor all components to the Composition API
* Switch stores from Vuex to Pinia


web/src/api/stores/file.store.ts
```diff
@@ -0,0 +1,118 @@
...
+  const apiStore = useApiStore();
+
+  // Hydrate the store
+  async function hydrate(): Promise<void> {
+    files.value = parse(await fetch(), apiStore.isAnonymous);
+    hydrated.value = true;
+  }
+
+  // Populate the ast & mapping of a given file.
+  async function populateFile(file: File): Promise<void> {
```
rien, line 102: Since the `ast` and `mapping` are always loaded (see line 55), this function can be removed? Or would it make sense to not load these values (since they require an extra `JSON.parse()` call)?

web/src/api/utils/cast.ts
```diff
@@ -0,0 +1,16 @@
...
+export function castToType(row: d3.DSVRowString): MetaRowType {
+  const rowValue = row.value;
+  const rowType = row.type;
+  const newRow = row as MetaRowType;
+  if (rowType === "boolean") {
+    newRow.value = rowValue ? rowValue.toLowerCase() === "true" : false;
+  } else if (rowValue && rowType === "number") {
+    newRow.value = Number.parseFloat(rowValue);
+  } else if (rowType === "object") {
+    newRow.value = null;
```
rien, line 13:
```suggestion
    // null values will have `rowType` "object"
    newRow.value = null;
```

maartenvn:
> ## General
> The pseudonymization checkbox seems to be clipped: image Would this be easily fixable?
> 

Fixed by 4f8ccbfa62b905e011200fa9e8567373b901bf41

> ## Overview page
> * The average similarity is reported as 0%, but this does not make sense. This may be a bug. In any case: I think the most logical way to calculate this, would be to take the average maximum similarity for each file.
> 

Fixed by a2e036787c496d0aa9411e00d665ffacb7fcd060

> ## Graph view
> Enabling pseudonymisation and enabling it again seems give console errors during the simulation.
> 

This is already fixed on another branch with performance improvements.

> ## File analysis
> This page does not seem to work (`TypeError: e is undefined` is printed to console), I think it might have something to do with lacking semantic data.

As discussed, this appears to be a bug in the CLI. Once boolean metadata values are fixed a fix can be applied.

maartenvn:
Closes #779

web/src/api/stores/file.store.ts
maartenvn: The performance impact of this parse action isn't really that big to justify fetching it on demand. I will remove this function & call.

rien (approved):

PR #786, opened 2022-06-30
Improve performance & fix graph bugs

maartenvn:
## Performance Improvements

This PR improves the performance & responsiveness of the web application.

### Data fetching & parsing

The initial data fetching & parsing has been improved:

| Before  |  After  | Improvement |
|---|---|---|
|  4868ms  | 2630ms |  45.97% faster |

This improvement is done by replacing most `ref`-values by `shallowRef`-values, which do not track changes on object properties. This is especially noticable when working with large data values.

### Graph view

The graph view does no longer render the initial ticks, when the nodes are still scattered. The calculations are still done but are not rendered by the browser. This also hides the "explosion" caused by the random scattering of nodes.

### Debounce cut-off value

The cut-off value of the similarity is now debounced by 100ms.

## Bug fixes

* Fixed issues with legend toggling not working correctly
* Fixed issue with labels having incorrect color mappings
* Fixed errors when switching to/from anonymous file names

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -109,17 +109,15 @@
 
     <v-main>
       <v-container class="container">
-        <keep-alive>
-          <router-view v-if="isLoaded" />
-          <loading v-else />
-        </keep-alive>
+        <router-view v-if="isLoaded" />
+        <loading v-else />
       </v-container>
     </v-main>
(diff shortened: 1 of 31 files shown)


rien (approved):
:rocket:

PR #787, opened 2022-06-30
Conditionally fetch semantic information, based on the metadata 

rien:
Instead of always trying to fetch `semantic.json` and catching a possible 404, this will now use information in the `metadata.csv`-file to conditionally fetch the semantic data.

This PR also changes how metada is written: 

Some values were confusingly serialised in the `metadata.csv` output:
- booleans would be written as an empty string for `false` and `1` for `true`
- null values would be written as an empty string

This could cause some weird behavior in the UI. 

Now, this output is written by performing `toString()` on the value if it is not null, or writing `null` otherwise.


maartenvn:
@rien This is ready 🙂

PR #818, opened 2022-08-03
Improve pair code viewer & add diff viewer 

### LINKED ISSUE #682: Better editor in pairwise view
Currently the pairwise view has issues with large files and uses complex extensions to prim.js. We may want to consider changing this editor with alternatives. Some options worth looking at:
- [CodeMirror](https://github.com/codemirror/CodeMirror)
- [Monaco](https://microsoft.github.io/monaco-editor/)

maartenvn:
# Improve pair view

## New code viewer

image

The pair view has been rewritten using `Monaco` as an editor instead of `Prism.JS`. The reasoning for using Monaco is that it offers a lot of features out of the box, requiring no plugins or large configuration. CodeMirror (which is used by Dodona) offers no native side-by-side diff view and requires more configuration to get things done. As the primary focus of the code viewer on this page is read-only, Monaco was chosen instead.

Matches are now shown in a less obtrusive blue and the current selected match in aqua-green.

## New diff view

A new diff view between the 2 files in a pair was introduced. This view is automatically selected when the similarity of a file is >= 80%, as the match view will show large matches at this similarity, making the diff view more interesting.

image

## Slight styling changes to the pairs table

image

### Note

When testing this branch on the SOCO dataset: matches in matches can be found (and will cause dark blue, less readable code patches). This is a bug in the lib and will be fixed in a seperate PR.

Closes #682


web/src/components/pair/PairCodeDiff.vue
```diff
@@ -0,0 +1,82 @@
...
+const initialize = (): void => {
+  // Monaco file models.
+  const leftFileModel = monaco.editor.createModel(props.pair.leftFile.content, props.metadata.language);
+  const rightFileModel = monaco.editor.createModel(props.pair.rightFile.content, props.metadata.language);
+
+  // Monaco diff editor
+  editor.value = monaco.editor.createDiffEditor(editorElem.value, {
+    enableSplitViewResizing: false, // Do not allow resizing of the diff view.
+    readOnly: true,
+    automaticLayout: true,
```
rien, line 46: Monaco's custom menu does not add anything of value I think, so could you add `contextmenu: false` here?

web/src/components/pair/PairCodeDiff.vue
```diff
@@ -0,0 +1,82 @@
...
+  });
+  editor.value.setModel({
+    original: leftFileModel,
+    modified: rightFileModel,
+  });
```
rien, line 51: Can you try if we can disable initally selecting the first line with `editor.setSelection(new monaco.Selection(0, 0, 0, 0));`?

web/src/components/pair/PairCodeMatchEditor.vue
```diff
@@ -0,0 +1,349 @@
...
+    value: file.value.content,
+    language: props.metadata.language,
+    readOnly: true,
+    smoothScrolling: true,
+    automaticLayout: true,
+    renderLineHighlight: "none",
+    renderValidationDecorations: "off",
+    minimap: {
+      enabled: false,
+    }
```
rien, line 216: The custom context menu can be disabled here as well.

web/src/components/pair/PairCodeMatchEditor.vue
```diff
@@ -0,0 +1,349 @@
...
+const editorElem = ref();
+// Monaco editor
+const editor = shallowRef<monaco.editor.IEditorOverrideServices>();
+// List of selections
+const selections = ref<Selection[]>([]);
+// Monaco editor decorations
+const decorations = shallowRef([]);
+
+// Get the match at a given editor position.
+const getMatchAtPosition = (row: number, col: number): Match | null => {
```
rien, line 80: It seems like some of the highlighted matches are "behind" others, making it not possible to select them. I think this may be fixed by trying to find the smallest match on the current position?

web/src/components/pair/PairCodeMatchEditor.vue
```diff
@@ -0,0 +1,349 @@
...
+    const nextMatch = matches.value?.[nextIndex];
+
+    // Scroll to the match.
+    scrollToMatch(nextMatch);
+    // Set the selected match.
+    selectedMatch.value = nextMatch ?? null;
+  });
+
+  // Allow using the shift+tab key to cycle through the matches, in reverse.
+  editor.value.addCommand(monaco.KeyCode.Shift | monaco.KeyCode.Tab, () => {
```
rien, line 267: This does not seem to work unfortunately.

rien (commented):
Nice work, the code is so much cleaner now :ok_hand:

maartenvn:
@rien The changes have been applied

rien (approved):

PR #821, opened 2022-08-05
Add lerna as monorepo manager

rien:


rien:
I'm closing this PR as it does not actually solve problems, and adds more complexity.

PR #822, opened 2022-08-05
Improve graph view

maartenvn:
# Improve graph view

image

## General UI cleanups

* Sliders & checkboxes now use Vuetify components instead of the native browser controls.
* Remove elevation in elevation
* Properly scale clustering table to page width
* Add some transitions
* Use correct font in tooltips
* Use correct cursor when element is clickable
* Lazy load tabs in the clustering card to improve performance

### Files list

The files list in the different tabs of the clustering card has been redesigned:
* Show header when scrolling (instead of scrolling entire card)
* Hide timestamp when not available
* Proper padding
* Remove elevation in elevation

| Before | After |
| ------- | ------ |
| image | image |

### Pairview

* The pairview is now an inline table instead of a link to a different view (which is bax UX in a tabview)

image

### Timechart

* General cleanup & some responsiveness

| Before | After |
| ------- | ------|
| image | image |

### Heatmap

* General cleanup & some responsiveness
* Use tooltip instead of alert for hovering
* Show highlighted tile using outline

| Before | After |
| ------- | ----- |
| image | image |

### Cluster

* General cleanup & some responsiveness

| Before | After |
| ------- | ------ |
| image | image |

## Pause button on graph

The graph now has a pause button to pause the simulation (thus stopping the movement of the nodes in the graph).

image

**Note**: pausing the simulation significantly improves the responsiveness of the page. In the future it may be interesting to pause the simulation automatically when a specific state has been reached.


maartenvn:
This PR is based on #818, which must be merged first in order to view the correct changeset.

rien (changes requested):
The file `src/components/settings/SingletonSetting.vue` is empty and does not seem to be used. Should it be removed?

rien (approved):

PR #823, opened 2022-08-05
On demand pseudonymization without refetch/parsing

### LINKED ISSUE #785: Perform pseudonymisation without downloading files again
Toggling the pseudonymisation now triggers a re-hydration of all the file stores, which should not be needed.

We can solve this by changing the file name and timestamp depending on the `isAnonymous` value.

maartenvn:
On demand pseudonymization without refetching the data & parsing the files.

The fix uses the `isLoaded` property to force re-rendering of the main view, which makes sure every component is updated to the new pseudo names. This is not the most performant solution but does not require changing `shallowRef` to `ref` in every component (which causes other performance issues).

Closes #785

### DIFF
--- a/web/src/api/stores/file.store.ts
+++ b/web/src/api/stores/file.store.ts
@@ -4,7 +4,7 @@ import { shallowRef, computed, nextTick } from "vue";
 import { DATA_URL } from "@/api";
 import { File, ObjMap } from "@/api/models";
 import { useApiStore } from "@/api/stores";
-import { colors, names, uniqueNamesGenerator } from "unique-names-generator";
+import { names, uniqueNamesGenerator } from "unique-names-generator";
 import { useLegend } from "@/composables";
 import { commonFilenamePrefix } from "../utils";
 
--- a/web/src/api/stores/file.store.ts
+++ b/web/src/api/stores/file.store.ts
@@ -51,9 +51,10 @@ export const useFileStore = defineStore("files", () => {
 
       // Store pseudo details.
       const pseudoName = randomNameGenerator();
+      const pseudoPath = `${pseudoName}.${filePathExtension}`;
       file.pseudo = {
-        path: `${pseudoName}.${filePathExtension}`,
-        shortPath: file.pseudo.path,
+        path: pseudoPath,
+        shortPath: pseudoPath,
         fullName: pseudoName,
         timestamp: extra.timestamp ? new Date(extra.timestamp.getTime() + timeOffset) : undefined,
         labels: String(labels.indexOf(extra.labels)) ?? "",
--- a/web/src/api/stores/file.store.ts
+++ b/web/src/api/stores/file.store.ts
@@ -25,7 +25,7 @@ export const useFileStore = defineStore("files", () => {
   // Parse the files from a CSV string.
   function parse(fileData: d3.DSVRowArray): ObjMap<File> {
     const randomNameGenerator = (): string => uniqueNamesGenerator({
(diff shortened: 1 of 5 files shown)


maartenvn:
This branch is based on #822 and requires merging this PR first.

maartenvn:
A red dot will be shown when pseudonymization is enabled.

image

web/src/api/stores/file.store.ts
```diff
@@ -41,34 +44,43 @@ export const useFileStore = defineStore("files", () => {
...
-        }
+      // Add the label to the labels list, if it doesn't exist yet.
+      if (!labels.includes(row.label)) {
+        labels.push(extra.labels);
       }
 
+      // Store pseudo details.
+      const pseudoName = randomNameGenerator();
+      file.pseudo = {
+        path: `/exercise/${pseudoName}.${filePathExtension}`,
```
rien, line 55: Is the (full) path shown in the UI?

If so, showing an anonymized version is not really useful. I would prefer a message instead, e.g. "Full path not available when anonymization is enabled."

web/src/api/stores/file.store.ts
```diff
@@ -23,16 +23,19 @@ export const useFileStore = defineStore("files", () => {
   const legend = useLegend(files);
 
   // Parse the files from a CSV string.
-  function parse(fileData: d3.DSVRowArray, anonymize = false): ObjMap<File> {
-    const randomNameGenerator = (): string =>
-      uniqueNamesGenerator({ dictionaries: [colors, names], length: 2 });
+  function parse(fileData: d3.DSVRowArray): ObjMap<File> {
+    const randomNameGenerator = (): string => uniqueNamesGenerator({
+      dictionaries: [colors, names],
```
rien, line 28: I would change this to only use the `names` dictionary, since colors in the name are a bit confusing. The length could also be just 1, since there are 4900 names which should be enough.

web/src/api/stores/file.store.ts
```diff
@@ -41,34 +44,43 @@ export const useFileStore = defineStore("files", () => {
...
+        labels: extra.labels,
+      };
+
       return [row.id, row];
     });
     const files: File[] = Object.fromEntries(filesMap);
 
     // Find the common path in the files.
-    const commonPath = commonFilenamePrefix(Object.values(files));
+    const commonPath = commonFilenamePrefix(Object.values(files), (f) => f.path);
```
rien, line 76: This function should really only be executed for the original files, since we determine the prefix of the `pseudo` object ourselves.

I also don't think the `pseudo`-objects need a prefix, because it does not have any meaning. (See the other comment about the full path).

web/src/api/utils/file.ts
```diff
@@ -13,18 +13,19 @@ export function fileToTokenizedFile(file: File): TokenizedFile {
 /**
  * Common filename prefix for a given list of files
  * @param files Files
+ * @param getPath Function to extract the path from the file
  * @returns Common prefix for all files.
  */
-export function commonFilenamePrefix(files: File[]): string {
+export function commonFilenamePrefix(files: File[], getPath: (f: File) => string): string {
```
rien, line 19: This change is not needed anymore if we only apply this function on the original file path.

rien (changes requested):

maartenvn:
@rien Requested changes have been applied

rien (commented):
There is an unused import warning and some documentation left.

rien (approved):

PR #841, opened 2022-08-10
Improve UI

maartenvn:
# Improve UI of the application

## Overview
- [x] Align text left, instead of centering

## File pairs
- [x] Similarity: show as percentage instead of fractional number. For example, with progress indicators:
image

## Similarity slider
- [x] Changing the slider lets the user lose the original position, mark this position as a tick (if possible) on the slider and enable a reset.

## Cluster view
- [x] Hide pagination of pairs table if less than 10 items.

## Graph
- [x] Hint nodes & clusters are clickable
- [x] Make nodes bigger & add more space between nodes in the cluster

## General
- [x] Check the colors for consistency (there is a difference between d3's blue and vuetify's blue for example) and contrast (d3's orange is not enough contrast with the background).


maartenvn:
I am working on a new overview page:

image

There is still a lot of white space on the first card on the left. If anyone has some ideas to put additional information or something else let me know.

@rien @bmesuere

maartenvn:
A reset button was added to the cut-off slider to allow reset to the initial value:

image

rien:
- The symbols in the three middle cards (similarity-cutoff, highest similarity, clusters) could be confused with buttons. I would avoid putting it in a box with the primary color. Instead, you could set the icon as a gray "background" of the card?
- I think listing the labels itself (possibly with their color) in addition to the amount, would be useful.
- I would maybe invert the information rows: `404 files` instead of `Files: 404`.

maartenvn:
> * The symbols in the three middle cards (similarity-cutoff, highest similarity, clusters) could be confused with buttons. I would avoid putting it in a box with the primary color. Instead, you could set the icon as a gray "background" of the card?

image

I changed the background color to a light blue. This should distinguish the card from a button.

maartenvn:
Improved UI for the selected cluster/node.

| Before | After |
| ------- | ------|
| image | image |

maartenvn:
image

Labels are now visible on the overview page.

rien:
These improvements are looking good :+1: 

##### Cluster
- If data is unavailable, I would omit the information. But there should be a nudge telling the user how they can add time-information next time. Maybe a timeline on the homepage that is disabled?
- Instead of "Author", I think using "Submission" is better, since it works for file names as well as student names.
- In addition, I don't know if we should show "Author:", "Label:", "Hand-in date:", I think the symbol is clear, and we could add a tooltip instead.
- I would use the colored circle again for the label, this will improve the recall.
- For "Selected cluster", I would invert the information:
  - **3** submissions in this cluster.
  - **100%** average similarity
 - If hand-in date is available, highlighting the first submitter in some way would be useful 

##### Overview
- I think it would look better to start the programming language with a capital letter.
- The label count can be useful as well "5 labels in this dataset"

maartenvn:
@rien This can be reviewed 🙂

web/src/d3-tools/GraphElementList.vue
```diff
@@ -1,55 +1,44 @@
...
-            </v-tooltip>
-          </td>
-        </tr>
-      </tbody>
-    </v-simple-table>
-  </v-card>
+  <v-simple-table class="graph-list" fixed-header>
+    <thead>
+      <tr>
+        <th>File</th>
```
rien, line 5:
```suggestion
        <th>Submission</th>
```

web/src/d3-tools/GraphElementListCard.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <v-card outlined>
+    <v-card-title> Files in this cluster </v-card-title>
```
rien, line 3:
```suggestion
    <v-card-title>Submissions in this cluster </v-card-title>

```

web/src/views/Overview.vue
```diff
@@ -1,76 +1,149 @@
...
+            <div class="stat-card-icon-background primary"></div>
+            <v-icon color="primary" x-large>mdi-chart-bell-curve</v-icon>
+          </div>
+
+          <div class="stat-card-content">
+            <h3 class="stat-card-title">
+              Highest similarity
+
+              <info-dot>
+                The highest similarity we've found between two files is {{ (highestSimilarity * 100).toFixed(0) }}%
```
rien, line 73:
```suggestion
                The highest similarity we've found between two submissions is {{ (highestSimilarity * 100).toFixed(0) }}%
```

web/src/views/Overview.vue
```diff
@@ -1,76 +1,149 @@
...
+            <div class="stat-card-icon-background primary"></div>
+            <v-icon color="primary" x-large>mdi-account-group-outline</v-icon>
+          </div>
+
+          <div class="stat-card-content">
+            <h3 class="stat-card-title">
+              Clusters
+
+              <info-dot>
+                Files are grouped into clusters based on their similarity.
```
rien, line 122:
```suggestion
                Submissions are grouped into clusters based on their similarity.
```

web/src/views/Overview.vue
```diff
@@ -1,76 +1,149 @@
...
+                If a file pair has a similarity above the threshold, they will belong to the same cluster.
```
rien, line 123:
```suggestion
                If a submission pair has a similarity above the threshold, they will belong to the same cluster.
```

web/src/views/Overview.vue
```diff
@@ -1,76 +1,149 @@
...
+              </info-dot>
+            </h3>
+            <div class="stat-card-value">{{ clustering.length }}</div>
+            <div class="stat-card-subtitle text--secondary">Largest cluster: {{ largestCluster }} files</div>
```
rien, line 127:
```suggestion
            <div class="stat-card-subtitle text--secondary">Largest cluster: {{ largestCluster }} submissions</div>
```

rien (changes requested):
The comparison page seems to be broken: the page does not load and the console spews out the error `r` is undefined.

There does also seem something weird happening on the cluster page: not all pairs are shown in the pairs table: a cluster with 7 submissions, only shows 6 pairs, bot there should be more.

web/src/components/pair/SimilarityDisplay.vue
```diff
@@ -0,0 +1,88 @@
...
+  progress: false,
+});
+
+// Convert the similarity value into a percentage.
+const value = computed(() => {
+  return (props.similarity.toFixed(2) * 100).toFixed(0);
+});
+
+// Determine the color of the similarity value.
+const color = computed(() => {
```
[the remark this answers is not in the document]
maartenvn: This is done in a later PR on a different branch

web/src/d3-tools/GraphElementList.vue
```diff
@@ -91,7 +81,7 @@ const hasTimestamp = computed(() => {
 
 const formatTime = (time?: Date): string => {
   if (!time) return "";
-  return DateTime.fromJSDate(time).toLocaleString();
+  return DateTime.fromJSDate(time).toFormat("dd/MM hh:mm");
```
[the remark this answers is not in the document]
maartenvn: These time displays are extracted in a seperate component in an upcoming PR and in order to prevent merge conflicts I will change this later on.

rien (approved):

PR #861, opened 2022-08-22
Submission overview and detail page

maartenvn:
# Submission overview and detail page

This PR adds an overview & detail page for the submissions in the processed dataset.

## Submission overview page

The submission overview page replaces the "File Analyis" page and offers a more compact overview without need for any pagination.

image

## Submission page

The submission page groups important information of a submission together on a comprehensive, shareable page. 

It contains:
* A table with other files and the corresponding similarity
* A cluster graph showing the cluster of the file
* A timeline showing the chronological order of submissions in the cluster
* A similarity & longest fragment graph, as previously available on the "File Analyis" page
* The code of the submission

image

## Sidebar

The sidebar navigation entries have been renamed to more comprehensive names and less emphasis is put on the pairs table.

image

## Other

* `/compare/:id` was renamed to `/pairs/:id` to be conform with the submissions. This will make breadcrumbs easier in the future.
* Removed old file analysis page (`/fileanalysis`)
* Removed unused about page component


web/src/components/SubmissionsTable.vue
```diff
@@ -0,0 +1,139 @@
...
+        </div>
+
+        <div class="submission-path text--secondary">
+          <v-icon small>mdi-file-document-outline</v-icon>
+          <span>{{ item.path }}</span>
+        </div>
+      </div>
+    </template>
+
+    <template #item.label="{ item }">
```
rien, line 28: Can the label row be hidden if there are no labels present in the dataset? I think it could be confusing to show "No Label", a user could expect some submissions to have a label.

web/src/components/SubmissionsTable.vue
```diff
@@ -0,0 +1,139 @@
...
+    @click:row="rowClicked"
+  >
+    <template #item.name="{ item }">
+      <div class="submission-info">
+        <div class="submission-name">
+          {{ item.name }}
+        </div>
+
+        <div class="submission-path text--secondary">
+          <v-icon small>mdi-file-document-outline</v-icon>
```
rien, line 22: I would only show the file path if the name is different from the (short)path.
We could also consider not showing the file path itself at all (since we try to guess the name based on the path anyway), or hiding it behind a tool tip, showing the full path.

web/src/components/SubmissionsTable.vue
```diff
@@ -0,0 +1,139 @@
...
+      <div class="submission-label">
+        <label-dot :label="item.label.label" :color="item.label.color" />
+        <label-text :label="item.label.label" colored />
+      </div>
+    </template>
+
+    <template #item.similarity="{ item }">
+      <span class="submission-similarity">
+        <similarity-display :similarity="item.similarity" progress dim-below-cutoff />
+      </span>
```
rien, line 38: An idea that popped in my mind, but I don't know how doable it is: showing which other submission(s) the current row has the highest similarity with (maybe only on hover?). What do you think?

web/src/components/FileTimestamp.vue
```diff
@@ -0,0 +1,48 @@
...
+  long?: boolean;
+}
+
+const props = withDefaults(defineProps<Props>(), {});
+
+const timestamp = computed(() =>
+  props.file?.extra?.timestamp ?? props.timestamp
+);
+const timeShort = computed(() =>
+  DateTime.fromJSDate(timestamp.value).toFormat("dd/MM hh:mm")
```
rien, line 36: I think this timestamp is still somewhat confusing, especially in locales where it is the habit to write `MM/dd`.

```suggestion
  DateTime.fromJSDate(timestamp.value).toLocaleString(DateTime.DATETIME_SHORT)
```

web/src/components/SubmissionsPairsTable.vue
```diff
@@ -0,0 +1,190 @@
...
+          </v-btn>
+        </template>
+
+        View submission
+      </v-tooltip>
+
+      <v-tooltip top>
+        <template #activator="{ on, attrs }">
+          <v-btn class="ml-2" icon small v-bind="attrs" v-on="on" :to="`/pairs/${item.id}`">
+            <v-icon>mdi-compare-horizontal</v-icon>
```
rien, line 78: I'm not convinced by this icon to go to the compare page. Maybe we can re-use the `file-compare` icon we're already using in the compare page itself?

web/src/components/clustering/TimeSeries.vue
```diff
@@ -183,3 +244,18 @@ onUnmounted(() => {
...
+
+<style lang="scss">
+.timeseries {
+  position: relative;
+
+  &-node {
+    cursor: v-bind("nodeCursor");
+
+    &.selected {
+      stroke: red;
```
rien, line 256: The red stroke is difficult to spot for nodes filled in red-ish colors, and it is also visually not great, clashing with the D3 color palette.
- We could use a black stroke to highlight the selected node, but this will be confusing with how we highlight the first submitter on the graph page.
- As an alternative, we can drop the saturation of other nodes, like we already do on the graph page. But this is not ideal for colorblindness.
- Having a concentric circle around it (stroke around an invisible larger circle), will be visually the best I think, but is technically a bit more cumbersome. This could either be achieved by using stroke gradients (hacky) or putting nodes in `g`roups in the svg.

web/src/components/SubmissionCard.vue
```diff
@@ -0,0 +1,56 @@
...
+    <v-row align="center" justify="space-between">
+      <v-col>
+        <v-card-title>
+          {{ props.file.extra.fullName ?? props.file.shortPath }}
+        </v-card-title>
+
+        <v-card-subtitle>
+          <div class="submission-card-info">
+            <v-icon :color="label.color" small>mdi-label-outline</v-icon>
+            <label-text :label="label.label" :color="label.color" colored />
```
rien, line 12: Hide the label when there are no labels in the dataset?

web/src/views/Submission.vue
```diff
@@ -0,0 +1,220 @@
...
+            </v-card>
+          </v-col>
+
+          <v-col cols="12" md="4">
+            <v-card>
+              <v-card-title>Information</v-card-title>
+              <v-card-text>
+                <div class="info-item">
+                  <v-icon :color="label.color">mdi-label-outline</v-icon>
+                  <label-text :label="label.label" :color="label.color" colored />
```
rien, line 67: Hide the label if there are no labels in the dataset?

web/src/views/Submission.vue
```diff
@@ -0,0 +1,220 @@
...
+            <v-card>
+              <v-card-title>Compare</v-card-title>
+              <v-card-subtitle class="pb-2">
+                Compare this submission with other submissions
+              </v-card-subtitle>
+
+              <submissions-pairs-table :file="file" />
+            </v-card>
+
+            <v-card class="mt-4">
```
rien, line 25: Should we hide the timeline if there is no timestamp data available?
If we would hide it, some users may be confused by it absence. So maybe we can add an information message on the homepage like "The dataset you analyzed did not contain timestamps, so some visualizations will not be available. Learn how to add metadata [here](https://dolos.ugent.be/guide/dodona.html)".

rien (changes requested):
Very good work, I think this page is a very useful addition :+1:

web/src/components/SubmissionsPairsTable.vue
maartenvn: I don't think this icon is suitable. It looks too much like a file and can be confusing with the submission.

I do think we should change the eye to same file icon (but with only one file instead of 2) as the submission page.

web/src/components/SubmissionsTable.vue
maartenvn: This is definitely possible but:
* This will be expensive, as all pairs matching a file must be found and sorted by similarity
* I would move this into a separate issue instead of patching it on this PR

web/src/components/clustering/TimeSeries.vue
maartenvn: Right now there is no first submitter highlighting on the graph, as this was an issue that needed some brainstorming.
I will change this to black for now.

web/src/views/Submission.vue
maartenvn: I would not remove the card, as it would be confusing. I will change this to the message you provided.

maartenvn: image

web/src/components/SubmissionsTable.vue
maartenvn: I would be consistent. If we don't show the file path it should be left out always.
I will add a tooltip showing the full path.

image

web/src/components/SubmissionsPairsTable.vue
rien: I don't think I agree that it looks too much like a file, it has a clear separation showing two files:
image

Another possible value is the `set-center` icon, since it is also used on the compare page:
image
Or, some kind of search icon, to signal closer inspection:
image
But that will also be confusing with showing the file itself.

The main reason why I think the current icon `compare-horizontal` doesn't really fit, is because we don't use it anywhere. It might as well mean swapping the current submission with that one (thus, essentially the same meaning as the other action). So it would maybe also help to re-use an icon on the pairs table so its meaning is consistent throughout the application.


For the eye: we can also make the whole row clickable, like we do with the other tables in the application.

maartenvn: From the options listed above `set-center` seems the best to me.

I will also switch to making the rows clickable, although I don't know if having a click action on top of a clickable row is very good UX?

maartenvn: Maybe as the row is now clickable we can replace the icon button by an actual button (as there is only one action left now):

image

rien (approved):

PR #864, opened 2022-08-23
Add breadcrumbs & back button

### LINKED ISSUE #852: Improve navigation: back-button and breadcrumbs
Navigation in the UI is currently not ideal, users lack a back button and can be confused how to go back in the UI. We can solve this by using breadcrumbs and back buttons. Some pages should also remember the state they were in when the user left.

maartenvn:
## Breadcrumbs & back button

Item specific pages now contain breadcrumbs & a back button to go back to the previous page. The breadcrumbs are only 1 deep and only return to the previous overview page.

image

## Maintain previous state

The value of the search fields is now encoded in the url, allowing easy sharing & better back behaviour. 

## Other changes

* Clusters on the graph are now encoded using a query param instead of a hash
* Submission cluster links are now linked to the corresponding cluster on the graph page
* Add `useRouteQuery` for encoding values into the querystring

Closes #852


maartenvn:
This requires merging & rebasing #861 first.

PR #865, opened 2022-08-23
Show more information on compare view

maartenvn:
During the usertests of the application some users found it hard to find which submission was submitted first when on the compare view.

More information has been added to this view:
* Author of a file (if available)
* Timestamp of the file (if available)

A button to go to the submission page was also added:

image


web/src/components/pair/PairCodeFileInfo.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <div class="file">
+    <div>
+      <div class="file-info-title">
+        <v-icon color="primary">mdi-file-outline</v-icon>
+        {{ props.file.shortPath }}
+      </div>
+
```
rien, line 8: I also think adding the label here (if any labels are present) would be useful here.

rien (approved):

web/src/components/pair/PairCodeFileInfo.vue
maartenvn: I will add this

maartenvn: Too prevent too much wrapping I put the extra information on a separate line:
image

PR #867, opened 2022-08-24
Fix comlink production errors on data worker

maartenvn:
Fix comlink production errors on data worker.

This is achieved by getting rid of `comlink-loader` and replacing it with manual `comlink` calls. This is a bit more verbose but still more convenient then working with regular workers.


rien (approved):
Works like a charm :rocket:

PR #868, opened 2022-08-25
Add global state settings

### LINKED ISSUE #853: Global state settings
We currently have some state that can be changed, but the controls are spread throughout the application. There should be a single location where all these settings (similarity threshold, hidden labels, anonymization) can be found and changed.

maartenvn:
This PR adds a global settings menu allowing to change:
* The similarity threshold
* If the dataset is anonymized
* If certain labels are enabled

image

The menu is opened using a gear icon on the top right of the navbar.

The behavior of the labels is now global, which means that toggling a label in a visualization or in the global settings will change the entire application to function as if the label was never included in the dataset in the first place.

### Other changes

* Rewrite of the label & legend system
* Change color 10 scale to 20 scale for datasets with more than 10 labels

Closes #853


web/src/api/stores/pair.store.ts
```diff
@@ -28,6 +28,47 @@ export const usePairStore = defineStore("pairs", () => {
...
+  function calculateActivePairs(): void {
+    // Return all files if no labels are available.
+    if (!fileStore.hasLabels) {
+      pairsActive.value = pairs.value;
+      return;
+    }
+    
+    const pairsFiltered = { ...pairs.value };
+
+    // Delete all fthe pairs that contain a file that shouldn't be displayed.
```
rien, line 54:
```suggestion
    // Delete all the pairs that contain a file that shouldn't be displayed.

```

web/src/components/clustering/FileTagList.vue
```diff
@@ -21,7 +21,8 @@ interface Props {
 }
 
 const props = withDefaults(defineProps<Props>(), {});
-const { legend } = storeToRefs(useFileStore());
+const fileStore = useFileStore();
+const { legend } = storeToRefs(fileStore);
```
rien, line 25:
```suggestion
```
It seems like this isn't used anymore.

web/src/components/overview/OverviewBarchart.vue
```diff
@@ -17,181 +15,162 @@ import { useElementSize } from "@vueuse/core";
...
+    .attr("transform", "translate(0, 0)")
+    .call(d3.axisLeft(xScale).tickFormat(d3.format(".0%")));
+  yAxis.append("text")
+    .text("Similarity")
+    .attr("font-size", 15)
+    .attr("fill", "black")
+    .attr("transform", `translate(-50, ${height.value / 2 + 35}) rotate(90)`);
+
+  // Add the data.
+  barchartContent
```
rien, line 114: Did we want to make the bars clickable?

rien (changes requested):

rien (approved):

PR #870, opened 2022-08-26
Allow passing a zip-file as input

rien:
Allows users to pass a zip-file as argument to Dolos.
The zip-file should contain a top-level `info.csv`-file listing the location of the desired files.

Dolos will unpack this zip-file (with `unzip`) to a temporary directory, perform the analysis on the files in this directory as if the `info.csv` was passed as input, and finally remove this temporary directory before showing the results.


maartenvn (approved):
👍

PR #871, opened 2022-08-26
Improve data visualisations

### LINKED ISSUE #855: Improve similarity distribution graph
- Origin in bottom left
- Try if other bin sizes would improve the visualization
- Add hints (more similar, more bad)
- Use color to show items above threshold
- Bins: show tooltip and make clickable
- ...

A brainstorm is needed first to gather a final design.

maartenvn:
## Improve similarity distribution histogram

* Highlight threshold using light/dark blue.
* Improve legibility of labels & use correct font
* Double bins from 10 to 20
* More space on homepage

image

## Improve similarity/longest fragment highlighted histogram

* Add axis labels
* Improve legibility of labels & use correct font
* Render hover line on top of bins
* Render hover line using dashes

image

## Improve timeline

* Add axis label
* Improve legibility of labels & use correct font

image

## Improve heatmap

* Prevent overflow on regular submission names.

## Improve tooltips

* Use same styling as Vuetify tooltips

image

## Misc

# Use switches instead of icons for labels  & remove column title

image

Closes #855


maartenvn:
@rien changes are starting from [b06ad93](https://github.com/dodona-edu/dolos/pull/871/commits/b06ad93b280b4b88cd7bba278bcff83ee626a137)

web/src/components/pair/SimilarityDisplay.vue
```diff
@@ -40,6 +40,7 @@ interface Props {
   progress?: boolean;
   dense?: boolean;
   dimBelowCutoff?: boolean;
+  showGreen?: boolean;
```
rien, line 43: This value is not used, and the locations where we'll use this `showGreen` will have the same functionality as `dimBelowCutoff`, so these props may be merged?

web/src/components/summary/PairStatHistogram.vue
```diff
@@ -126,12 +127,15 @@ const histogramContent = histogramChart
   .append("g")
   .attr("transform", "translate(" + margin.left + "," + margin.top + ")");
 
+// Tooltip
+const tooltip = useD3Tooltip({ relativeToMouse: true });
+const tooltipMessage = (d): string => {
+  return `There are <b>${d.length}</b> files that have a value between <b>${d.x0}</b> and <b>${d.x1}</b>`;
```
rien, line 133:
```suggestion
  return `There are <b>${d.length}</b> submissions that have a highest similarity between <b>${d.x0}</b> and <b>${d.x1}</b>`;
```

web/src/components/summary/PairStatHistogram.vue
```diff
@@ -142,99 +146,112 @@ const draw = (): void => {
...
-    })
-    .attr("width", (d) => {
-      return Math.max(xScale(d.x1 || 0) - xScale(d.x0 || 0) - 1, 0);
-    })
-    .attr("height", (d) => {
-      return height.value - yScale(d.length);
-    })
-    .style("fill", d => getBinColor(d));
+  // Only add the data if there are any files available.
+  if (maxFileData.value.length > 0) {
```
rien, line 195: This chart is very similar to the similarity distribution chart shown on the overview page. Do you think it would make sense to extract some of the functionality there to a common component or function?

web/src/views/Overview.vue
```diff
@@ -101,13 +101,13 @@
               </info-dot>
             </h3>
             <div class="stat-card-value">{{ clustering.length }}</div>
-            <div class="stat-card-subtitle text--secondary">Largest cluster: {{ largestCluster }} submissions</div>
+            <div class="stat-card-subtitle text--secondary">Based on current threshold</div>
```
rien, line 104: This may be too long:
```suggestion
            <div class="stat-card-subtitle text--secondary">File pairs with a similarity above the current threshold are grouped in the same cluster.</div>
```
If so, then the current text is probably good.

rien (changes requested):

web/src/components/pair/SimilarityDisplay.vue
maartenvn: This prop is unused, so it can be removed

web/src/views/Overview.vue
maartenvn: This is unfortunately too long: 
image

web/src/components/summary/PairStatHistogram.vue
maartenvn: I am putting this into a separate issue, together with some cleanup to the other d3 components (as they also have pieces that can be more generic)

PR #879, opened 2022-08-30
Fix clustering algorithm on threshold of 100%

### LINKED ISSUE #848: No graph hull in cluster view when cut-off is 100%
No graph hull in cluster view when cut-off is 100%

image

maartenvn:
Fix clustering algorithm on threshold of 100%

Closes #848


rien (approved):
:+1:

PR #884, opened 2022-08-31
Upload page & support for multiple datasets

maartenvn:
This PR adds support for running the Web UI as front-end for the backend API with support for viewing multiple reports and uploading a zip to analyze.

Closes https://github.com/dodona-edu/dolos/issues/1086


maartenvn:
The note about the storage and sharing of the data will now be shown during the upload & analysis:

image

web/.env
```diff
@@ -1,3 +1,3 @@
 VUE_APP_API_URL=http://localhost:3000
 # Enable the line below to develop for server mode
-# VUE_APP_MODE=server
+VUE_APP_MODE=server
```
rien, line 3: I have added a shortcut for developing for the server: `yarn serve:server` and `yarn build:server`.

web/src/components/upload/UploadFormCard.vue
```diff
@@ -0,0 +1,494 @@
...
+        }
+      }
+    } catch (e: any) {
+      //
+    }
+  }, pollingInterval);
+
+  // Add the report.
+  pollingReports.value.push(reportId);
+};
```
rien, line 455: You've already split the original file up into multiple parts, which is good. But this component is still very large.

Would it make sense to put some of the logic here in a store?

web/src/components/upload/UploadsTableDeleteDialog.vue
```diff
@@ -0,0 +1,106 @@
...
+const loading = ref(false);
+const snackbar = useSnackbar();
+
+// Cancel the deletion.
+const cancel = (): void => {
+  open.value = false;
+};
+
+// Confirm the deletion.
+const confirm = async (): Promise<void> => {
```
rien, line 29: I think functions like this are maybe better collected in a `report.store.ts`. Deleting a report could be something we also want to do from other places in the application as well.

maartenvn:
Added a table containing all analysed reports. Reports are stored in local storage, as we don't offer authentication/user accounts yet.

image

maartenvn:
Dolos will now also show extra options like sharing and more information about the report when in server mode:

image

maartenvn:
Added a download button to download the current dataset:

image

Closes #1086

web/src/components/upload/UploadsTableDeleteDialog.vue
```diff
@@ -31,14 +33,10 @@ const confirm = async (): Promise<void> => {
 
   try {
     // Attempt to delete the upload.
-    await axios.delete(
-      `${process.env.VUE_APP_API_URL}/reports/${props.report.id}`
-    );
+    await axios.delete(appmode.reportUrl.value);
```
rien, line 36: If uploading a dataset works, but creating a report fails (because of an error), that report will have a local `error` status without associated report ID. Trying to delete this report here will however also fail (because the id is `undefined` and the report does not exist on the server).

I think if there is no associated report id, we can just delete it locally.

In addition, maybe we should automatically delete failed reports if they are older than a day?

web/src/components/upload/UploadsTableShareDialog.vue
```diff
@@ -0,0 +1,84 @@
...
+        <v-text-field
+          v-model="shareUrl"
+          label="Share URL"
+          outlined
+          dense
+          readonly
+          @focus="$event.target.select()"
+        />
+
+        <v-alert type="info" text>
```
rien, line 71: Can you move this above the text field? I think the copy button should be as close to the URL itself as possible.

web/src/components/upload/UploadsTable.vue
```diff
@@ -0,0 +1,133 @@
...
+  { text: "Upload date", value: "date", sortable: true },
+  { text: "Status", value: "status", sortable: true },
+  { text: "", value: "actions", sortable: false, align: "right" },
+]);
+
+// Table items
+// In the format for the Vuetify data-table.
+const items = computed(() =>
+  reports.reports.map((report) => ({
+    name: report.name,
```
rien, line 33: Maybe we should also show the local reference next to the report name, in case multiple datasets with the same name were uploaded?

rien (changes requested):
Looking good :+1: 

- There seems to be a bug here where the share URL does not work and throws `Error: Not Found` in the console.
- Deleting a report also does not work here (even successful ones), maybe this is because I have a report that has an error that cannot be deleted.
- I also get some warnings in the console when opening a report (`Property or nethod "submissionsCount" is not defined in Overview.vue`)

web/src/components/upload/UploadFormCard.vue
maartenvn: This was later implemented in the "report store"

web/src/components/upload/UploadsTableShareDialog.vue
maartenvn: image

web/src/components/upload/UploadsTable.vue
maartenvn: I am not sure about this yet, as the table will become too cramped. I was under the impression the date would be enough of an indicator to distinguish the different datasets?

rien (approved):

PR #907, opened 2022-10-17
Update documentation to reflect UI update

rien:
The website has been updated in preparation for a new release:
- Added the demo datasets
- Updated with more images, replaced old images with new ones
- Mention performing an analysis using `info.csv` files and export ZIP-archives

The updated version is already deployed to https://dolos.ugent.be/.


pdawyndt:
Some observations

- The two uses cases on the introduction page that are illustrated with the image from the Dodona manuscript have some text that mentions "test" and "mandatory exercises" that are explained in the manuscript but come out of nowhere in the introduction section. I think it would be better to describe the two uses cases in more general terms: contexts where no communication or collaboration is allowed and thus no similar submissions are expected, and contexts where students are allowed to collaborate in small groups and where we want to see if solutions are not exchanged among larger groups of students.

Some ideas

- Make the plagiarism graph on the home page clickable (directly leads to the live example of this plagiarism graph)
- As we have for [Unipept](https://unipept.ugent.be/publications), we might also add a Publications page for Dolos. Maybe in addition to the publications that directly focus on Dolos, we could also mention the other Dodona articles, but clearly with some priority between the two.

rien:
The long loading times were caused by the original site deploying over itself, this should be fixed now.

rien:
- To make the plagiarism graph clickable, an update to the Vuepress theme is required. We can enable this when this PR is merged: https://github.com/vuepress/vuepress-next/pull/1136
- A publication page is a good idea, but I'll postpone this for a future update (#909)

PR #1039, opened 2023-01-30
Add pairs count to similarity distribution histogram

rien:
In order to get a better feeling for how to improve the similarity threshold, I've edited the similarity distribution histogram to also show pairs.

Pairs are added as very faint blue bars.
The scale is changed to a log scale because of the high number of pairs.

**Do we want to provide this visualization in the web UI?** This can be behind a toggle button (Include pairs) or in a separate view (which could include extra statistics about the dataset results).

In addition, how can we improve this visualization? I see the following problems:
- Showing pairs as faint blue bars might be confusing, should another color be used?
- What title to put on y-axis? "Similarity frequency"?
- Should the log axis be toggleable?

Below some examples of this change:

## Pyramidal constants - exercise

Without:
image

With pairs:
image

## Pyramidal constants - evaluation

Without:
image

With pairs:
image

## Plutokiller - exercise

Without:
image

With pairs:
image

## Highly similar exercise (R1)

Without:
image

With pairs:
image

## Highly similar exercise (R2)

Without:
image

With pairs:
image

This last case adds a lot of "explaantion" as to why the threshold is chosen at 25%: it is is the "first" local minimum that Dolos finds. How to improve this threshold selection is for another discussion.

### DIFF
--- a/web/src/components/overview/OverviewBarchart.vue
+++ b/web/src/components/overview/OverviewBarchart.vue
@@ -10,7 +10,7 @@ import {
   onMounted,
 } from "vue";
 import { storeToRefs } from "pinia";
-import { useApiStore, useFileStore } from "@/api/stores";
+import { useApiStore, useFileStore, usePairStore } from "@/api/stores";
 import { useElementSize } from "@vueuse/core";
 import { useD3Tooltip, useRouter } from "@/composables";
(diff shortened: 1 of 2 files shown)


rien:
As we've discused: histograms with a log axis are not good, and this histogram does not add enough value.

PR #1076, opened 2023-03-15
Add a name and date to reports and show them in the Web UI

rien:
We've noticed that reports were not easily distinguished from each other, so we've changed Dolos to add names to each generated report and show them in the Web UI. Dolos will try to generate a fitting name based on the analyzed files, or the usr can pass a custom name. Additionally, we also record the report's creation date to avoid any further confusion if a dataset is analyzed multiple times.

This PR introduces the following changes:
- Dolos CLI accepts a new argument `-n, --name <name>` which allows the user to specify the name for a report
- Dolos' `metadata.csv` will now include two extra rows `reportName` and `createdAt`. The `reportedName` will be based on the analyzed files, or use the user specified name if given.
- The Dolos Web UI will show the report name on every page, as well as set the page title. This allows users to easily find if the correct report if multiple reports are opened in the browser.

image

A preview is available at https://dolos.ugent.be/demo/pyramidal-constants/exercise/branch/feature/report-names/#/


rien:
I currently show the report name on a few locations in the Web UI:

image

- In the HTML title
- In the report URL
- In the top app bar, next to the brand title. This is shown on every page.
- As the main `h1` title on the overview page where it would previously mention "Dolos".

I'm not 100% sure about this design. In #884, we added the report name to the information card like this:
image

But I think the title of the page "DOLOS" is not necessarily needed and the report name would be a better opener. If no report name is available, we can still fall back to "DOLOS".

Any opinions? @bmesuere @maartenvn @pdawyndt

pdawyndt:
I would opt for a breadcrumbs-like approach in the navbar as we do in Dodona

Dolos <project> <specific page>

- Dolos: linked to home page (where you can see listing of all recent analysis or launch a new analysis)
- <project>: name of current project linked to landing page with overview of project
- <specific page>: some view of the project (e.g. pairwise view of two submissions)

rien:
> I would opt for a breadcrumbs-like approach in the navbar as we do in Dodona
> 

We already have breadcrumbs, but we do not use the top navbar for it:

image

I think changing the breadcrumbs to use the top navbar is not ideal. So I would stay with the contents of the first screenshot I posted?

bmesuere (approved):
Showing it in the header is ok for me, but maybe using an n-dash, m-dash or other symbol to put between Dolos and the name looks better?

Was the huge file added by accident?

rien:
> Was the huge file added by accident?

No, the huge file(s) are the demo dataset which I've added as default dataset. This dataset is much more useful than the sample javascript dataset.

PR #1079, opened 2023-03-23
Improve graph performance and code quality

rien:
This PR is a large rewrite of the D3 force graph implementation of the _plagiarism graph_.

We now use the HTML `canvas` to render the Graph. The canvas more efficient in rendering many nodes and edges, where the previous SVG-implementation would struggle. This also solves a bug where the cluster SVG-elements were removed and added each simulation tick, which greatly decreased perfomance.

The underlying simulation has been tweaked to cool down faster. This also fixes a bug where in some situations, the simulation would not cool down at all.

The code quality has been improved: the D3 force graph is now a _composable_ which hides complexity. This composable is also split into multiple code files with clearly defined interactions and responsibilities instead of a large 600+ LOC component.

image

You can try out this PR live at https://dolos.ugent.be/demo/pyramidal-constants/exercise/branch/bug/graph-performance/#/graph
(any demo should work if you append `branch/bug/graph-performance` to the URL).


pdawyndt:
If you hide the nodes for all labels, and then display one of the labels again, nothing is displayed. Only when you display a second label (and sometimes even only when a third label is displayed) are the nodes of all the previous labels displayed in the graph.

pdawyndt:
Separation between nodes could be increased for clarity of the graph.

bmesuere (changes requested):
The speedup is amazing!

I don't know if it's possible, but the "reheating" start on click, which is a  bit strange. Is it possible to only do this on drag? I agree with Peter that the spacing between the nodes might be a bit larger.

Can you choose the "enter" animation for the cards?

rien:
- Forces have been tweaked to have a better distribution, the graph will also better react to resizing
- Some bugs have been fixed (notably #1085)
- The enter animation of cards is now a smoother slide-down

bmesuere (approved):

PR #1090, opened 2023-04-14
Add 'View pair' button to graph when applicable

### LINKED ISSUE #1015: More direct access to pairwise view
At some places it takes me some extra clicks to end up with a pairwise view, where I would expect a more direct access.

For example, when selecting a cluster of two in the graph view (= common case), we could add a direct link to the pairwise view in the panel.

image

maartenvn:
Add 'View pair' button to graph when only 2 submissions are in a single cluster.

image

Closes #1015

### DIFF
--- a/web/src/api/stores/pair.store.ts
+++ b/web/src/api/stores/pair.store.ts
@@ -116,7 +116,7 @@ export const usePairStore = defineStore("pairs", () => {
     return pairsActiveById.value[id];
   }
 
-  // Get a list of pairs for a given file.
+  // Get a list of pairs for a given list of files.
   function getPairs(file: File): Pair[] {
     return pairsActiveList.value.filter((pair) => pair.leftFile === file || pair.rightFile === file) ?? [];
   }
--- a/web/src/api/stores/pair.store.ts
+++ b/web/src/api/stores/pair.store.ts
@@ -116,7 +116,7 @@ export const usePairStore = defineStore("pairs", () => {
     return pairsActiveById.value[id];
   }
 
-  // Get a list of pairs for a given list of files.
+  // Get a list of pairs for a given file.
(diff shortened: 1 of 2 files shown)


web/src/api/stores/pair.store.ts
```diff
@@ -117,7 +117,7 @@ export const usePairStore = defineStore("pairs", () => {
     return pairsActiveById.value[id];
   }
 
-  // Get a list of pairs for a given file.
+  // Get a list of pairs for a given list of files.
```
rien, line 120: This does not seem to be correct?

web/src/components/graph/GraphSelectedInfo.vue
```diff
@@ -94,6 +101,7 @@ interface Props {
 }
 
 const props = withDefaults(defineProps<Props>(), {});
+const pairStore = usePairStore();
```
rien, line 104:
```suggestion
```

web/src/components/graph/GraphSelectedInfo.vue
```diff
@@ -85,6 +91,7 @@ import { File, Legend } from "@/api/models";
 import { Cluster, Clustering } from "@/util/clustering-algorithms/ClusterTypes";
 import { DateTime } from "luxon";
 import { getClusterElements } from "@/util/clustering-algorithms/ClusterFunctions";
+import { usePairStore } from "@/api/stores";
```
rien, line 94:
```suggestion
```

web/src/api/stores/pair.store.ts
maartenvn: This is a mistake, will correct this.

maartenvn:
Requested changes have been applied.

rien (approved):

PR #1111, opened 2023-04-28
Add archiving system with uploads

### LINKED ISSUE #1108: Dolos server: handle reports deleted on the server side
Currently, when a report is (automatically) deleted server-side, the client still thinks this report is present. Trying to do anything with this report will report in 404's and even trying to delete it will result in an error.

We should fix this by checking whether the report is still available and, if it's not, we should mark a report as "deleted" or "archived" (maybe "unavailable"?).

I also think deleting a report from localstorage should always work, even if trying to delete it from the server failed.

maartenvn:
### New archiving system
A new archiving system has been added. When visiting the upload page the status for all finished reports will be fetched from the API. If the report is no longer available, it will be marked as "Archived".

Archived reports will no longer be able to be viewed and can be safely removed from local storage without errors.

image
image

### Other changes
- Router syntax has been improved using [nested routes](https://router.vuejs.org/guide/essentials/nested-routes.html)
- Errors are now properly displayed, rather than keep loading forever.

### NOTE

This branch is based on https://github.com/dodona-edu/dolos/pull/1076, which must be merged first.

Closes #1108


rien (approved):
:+1: 

The nested routes are a much better solution. Nice!

rien (changes requested):
The status "archived" implies the report can still be recovered, which is not the case.

I think "deleted" is a better name for that situation, with the notice
> This report has been deleted on the server and is no longer available. Old reports can be deleted periodically to save storage space.

rien (approved):

PR #1153, opened 2023-06-21
Refactor to Vue 3 & Vuetify 3

maartenvn:
This branch updates and refactors the `dolos-web` project to use Vue 3, Vuetify 3 and Vite.

In addition, this adds a new project `dolos-core` which contains all standalone javascript / typescript code (not dependent on Node) from `dolos-lib` (see #1189). This allows `dolos-web` to directly depend on `dolos-core` without having to filter out Node-specific dependencies.

Finally, all projects have been migrated to use ES Modules (closes #574).

This branch can be previewed on https://dolos.ugent.be/demo/pyramidal-constants/exercise/branch/refactor/vue-3/#/

Closes #754

PR #1188, opened 2023-08-01
Switch to ESM modules

rien:


rien:
This change is already merged in #1153

PR #1387, opened 2024-02-27
Fix some issues with the docker-compose.yml

rien:
This PR:
- Adds missing `web/Dockerfile` and `api/docker-compose.yml` files
- Downgrades the API ruby version to 3.2 due to some issues on aarch64
- Locks the mariadb version to 11 and sets the collaction to `utf8mb4_general_ci` to avoid collation issues


rien:
@mikaelGusse can you possibly try out if this PR fixes your issues?

mikaelGusse:
Looks like it did fix the issues. I initially had an issues with this commit that the dockerfile's 
FROM ruby:3.2 installed version 3.2.3 which caused a conflict with the gemfile but changing that to 3.2.2 fixed that issue. Now it builds and runs with docker-compose correctly. Thanks!

I do still have a question. How do I initiate the analysis to start on these datasets in the api website?
image
They are queued there but the analysis does not appear to start automatically. Is this intended behaviour?

rien:
Good to here it is starting now. Thanks again for reporting and going through the hassle of testing 🙂

If you are using the docker-compose, it should start a worker which would quite instantly pick up the job from the queue. Can you confirm.the worker service is up? If so, do you see anything in the logs about the worker?

mikaelGusse:
The api seems to be sending these sql queries constantly in the terminal:
image

The worker appears to not be functional though:
image

image

Do i need to configure it somehow?

rien:
@mikaelGusse some environment variables were missing, I've added them and updated to ruby 3.2.3

It is possible you have to run `docker pull ghcr.io/dodona-edu/dolos:latest` as well, as our worker spawns new containers with this image to run the analysis.

I hope that is everything, let me know if something is still not working.

mikaelGusse:
Looks like it is now working perfectly. Thank you for the help!

rien:
You're welcome. Glad we could help!

Feel free to contact us again if you have feedback about Dolos, or need some information about integration with other services.

PR #1424, opened 2024-03-13
Update Dolos API for streamlined automated submissions

rien:
- We've enabled all origins for CORS, allowing all any application to interact with the API
- Removed tokens from the API and database, as they were not used anywhere
- Added `name` to the `dataset` and `report` JSON representations
- Added `html_url` to the `report` JSON representation, pointing to the front-end URL visualizing the report


bmesuere:
Thanks @rien!

Issue #1518, opened 2024-05-07
Self hosting Dolos using docker-compose


mikaelGusse:
**Which component(s) is your question about?**
Self-hosting Dolos web app

**What is your question?**
Hello! We here at Aalto are having some difficulties hosting Dolos on our own servers at the moment. We are trying to use the docker-compose method to host it. We are using nginx as well for this setup. Currently we are able to open up the web front end and submit a file for analysis. It manages to complete the analysis succesfully (most of the time) but when we try to open up the report it tries to fetch the report from the wrong place for some reason. Instead of going through https://DOLOS_URL/api/reports it tries to get the report from http://DOLOS_URL/reports. This happens despite us settings the VITE_API_URL to the https. 

Our current nginx.conf looks like this:
```
events {}

http {

    upstream dolos-api {
        server api:3000;
    }

    upstream dolos-web {
        server web:8080;
    }

    server {
        listen 80;
        server_name {INSERT DOLOS URL HERE};

        location / {
            return 301 https://$host$request_uri;
        }
    }

    server {
        listen 443 ssl;
        server_name {INSERT DOLOS URL HERE};

        ssl_certificate /etc/ssl/private/fullchain.pem;
        ssl_certificate_key /etc/ssl/private/privkey.pem;

        location /{
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-Host $host;
            proxy_pass http://dolos-web/;
        }

        location /api {
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-Host $host;
            proxy_pass http://dolos-api/;
        }
    }
}

```

And our current docker-compose looks like this

```
name: dolos
version: "3.9"
services:
  db:
    image: mariadb:11
    volumes:
      - dolos-db-data:/var/lib/mysql
    environment:
      MYSQL_ROOT_PASSWORD: {INSERT PASSWORD HERE}
    networks:
      - dolos
    healthcheck:
      start_period: 2s
      interval: 5s
      test: healthcheck.sh --su-mysql --connect --innodb_initialized
  web:
    image: ghcr.io/dodona-edu/dolos-web
    build: ./web/
    environment:
      VITE_HOST: 0.0.0.0
      VITE_PORT: 8080
      VITE_API_URL: https://{INSERT DOLOS URL HERE}/api
      VITE_MODE: server
    depends_on:
      - api
    networks:
      - dolos
    healthcheck:
      start_period: 2s
      interval: 5s
[...]

PR #1522, opened 2024-05-15
Fix hosting Dolos publicly on https using docker-compose

rien:
If Dolos is hosted publicly behind a reverse proxy, Rails does not appropriately detect that HTTPS should be used for urls generated by the API, causing the front-end to fetch the CSV files using http instead of https (see #1518).

This PR fixes this issue by introducing a new environment variable `DOLOS_API_URL` which is used to detect the hosting properties. It supersedes the environment variables `DOLOS_API_HOSTS` and `DOLOS_API_DISABLE_FORCE_SSL`.

Unfortunately there is a bug in rails where files served using ActiveStorage don't listen to the `relative_root_url` (https://github.com/rails/rails/issues/43633). This can cause issues if the API is hosted on a subdirectory (`https://example.com/api`). You can avoid this issue by using another application/proxy to set the base url (like Phusion Passenger's [`passenger_base_uri`](https://www.phusionpassenger.com/docs/references/config_reference/nginx/#passenger_base_uri)), or by adding an extra redirect location from `/rails/*` to `/rails/api/*`.

For example for NGINX, the following snippet would host Dolos on `https://dolos.localhost:9090`:

```nginx
http {
    upstream dolos-api {
        server api:3000;
    }

    server {
        listen 9090 ssl;
        server_name dolos.localhost;
        ssl_certificate /root/ssl/cert.crt;
        ssl_certificate_key /root/ssl/cert.key;


		# Serve the static front-end files
        location / {
            root /path/to/dolos-web;
        }

        location /api {
            proxy_set_header Host $host;
            proxy_set_header X-Forwarded-Host 'dolos.localhost:9090';
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Proto $scheme;
            proxy_pass http://dolos-api/;
        }

		# Required due to a bug in Rails' ActiveStorage
        location /rails {
            return 302 https://dolos.localhost:9090/api$request_uri;
        }
    }
}
```

### DIFF
--- a/api/config/environments/production.rb
+++ b/api/config/environments/production.rb
@@ -1,9 +1,12 @@
 require "active_support/core_ext/integer/time"
 
 Rails.application.configure do
-  config.hosts = ENV.fetch('DOLOS_API_HOSTS') { 'dolos.ugent.be' }
 
+  config.dolos_api_url = URI.parse(ENV.fetch('DOLOS_API_URL') { 'https://dolos.ugent.be/api' })
   config.front_end_base_url = ENV.fetch('DOLOS_API_FRONT_END_URL') { 'https://dolos.ugent.be/server' }
+
+  config.hosts = ENV.fetch('DOLOS_API_HOSTS') { config.dolos_api_url.host }
+
 
   # Settings specified here will take precedence over those in config/application.rb.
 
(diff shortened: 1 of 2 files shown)


docs/docs/hosting-dolos.md
```diff
@@ -60,7 +60,14 @@ Environment variables of the `dolos-web` service (the front-end web UI):
...
-- `DOLOS_API_FRONT_END_URL`: full URL (format: `https://{hostname}:{port}`) where the front-end is hosted
-- `DOLOS_API_HOSTS`: hostname and port of the API (format: `{hostname}:{port}`)
-- `DOLOS_API_DISABLE_FORCE_SSL`: disables https when set to `true`. **Warning:** always use https when publicly hosting dolos.
+- `DOLOS_API_URL`: full URL where Dolos API will be hosted, including the subdirectory (e.g. `http://localhost:3000/api`)
+- `DOLOS_API_FRONT_END_URL`: full URL where the front-end is hosted (format: `https://{hostname}:{port}`)
 - `DOLOS_API_DATABASE_{HOST,USERNAME,PASSWORD}`: hostname, username and password of the database server (MariaDB or MySQL)
+
+::: info
+
+When hosting the Dolos API **on a subdireectory** (for example: `/api/`) using Docker behind a reverse proxy (**nginx**, **apache**, **haproxy**), there is currently a bug where Rails is generating incorrect URL's not respecting this subdirectory.
```
pdawyndt, line 69: subdireectory → subdirectory

PR #1596, opened 2024-09-26
Update docker-compose.yml

chouenyu0808:
Fixed the problem that docker was not rebuilt when the api service changed IP.


chouenyu0808:
Fixed the problem that docker was not rebuilt when the api service changed IP.

PR #1597, opened 2024-09-26
Update docker-compose.yml

chouenyu0808:
Fixed the problem that docker was not rebuilt when the api service changed IP.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,6 +32,7 @@ services:
       start_period: 2s
       interval: 5s
       test: curl --fail http://localhost:8080 || exit 1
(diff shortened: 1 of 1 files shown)


chouenyu0808:
Fixed the problem that docker was not rebuilt when the api service changed IP.

chouenyu0808:
Fixed the problem that docker was not rebuilt when the api service changed IP.

rien:
Hi @chouenyu0808, there is already a similar fix on master that adds this feature to the dolos-web Dockerfile: https://github.com/dodona-edu/dolos/blob/main/web/Dockerfile#L20

This is currently not released however, so you need to build the dolos-web docker yourself currently.

chouenyu0808:
Can you tell me the steps to use it? Because I still can't use it

Rien ***@***.***> 於 2024年9月26日 週四 下午3:42寫道：

Issue #1588, opened 2024-09-19
Improve comparison by highlighting code fragments better


rien:
In #1583 it was mentioned that the comparison could be a bit clearer. It is not clear which highlighted parts are matched with each other. If the user clicks on a highlighted/matched fragment where the other fragment is out of the view, it scrolls to that part, but this is not obvious and bugged.

![Screenshot of the Dolos UI comparing two similar submissions](https://github.com/user-attachments/assets/59b16adf-a4dc-4457-8b6d-73a6393de0ce)

### Possible improvements
- Show matched fragments that are **ignored** because of template code, but visualize them in a distinct manner (e.g. greyed-out?)
- Add a visible way to highlight matched fragments, for example a label or tag in the top of each fragment
- A way to iterate or scroll through the matching Fragments
- A way to  show the underlying matching tokens contained within a match

In the past there has been a a collapsible sidebar with a list of all fragments, this was removed because the UI was confusing and complex. We could re-implement this with a better UI.

### What should be changed

The code responsible for the highlighting is contained within the components in https://github.com/dodona-edu/dolos/tree/main/web/src/components/pair - specifically `PairCodeMatchEditor.vue` (one side of the pair) and `PairCodeMatch.vue` (both sides, with communication between).

The code quality in these components is not superb because it mostly comes down to convincing the [Monaco Editor](https://microsoft.github.io/monaco-editor/) to show these fragments properly, and highlighting sections of code is a tricky thing.
While adding this feature, these components could be split up and cleaned up a bit.

PR #1600, opened 2024-09-27
Fix code highlighting bug

mikaelGusse:
Currently in Dolos the feature to highlight which fragments are matched in the Matches view is broken. This is due to a small bug in the corresponding code which checks if the hovering/selected match is equal to the match whose decoration we are supposed to be changing. 

This PR changes how the equivalence is checked from the JS === to a small helper function called "areMatchesEqual" which checks that two matches have the same values. 

Feel free to suggest other solutions to this bug, but this seems to work quite well. I also changed the alpha value of the selected match to be less transparent.

Related issue: #1588


rien (approved):
It seems to work now, thanks!

PR #1603, opened 2024-10-02
Change API docs to allow custom URL's

rien:
This adds an input field to the API documentation which allows someone to change the endpoint URLs shown.

This change is already live and can be viewed on https://dolos.ugent.be/docs/api.html

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -2,11 +2,70 @@
 outline: [2,3]
 ---
 
+<script setup>
+import { ref } from 'vue';
+const defaultURL = "https://dolos.ugent.be/api";
+const localhostURL = "http://localhost:3000";
+const dolosURL = ref("https://dolos.ugent.be/api");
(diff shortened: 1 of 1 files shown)


mikaelGusse:
I really like this idea! Although I am not sure why but the buttons for public API and localhost API don't appear to be working on my side. Tried on Firefox and Chrome on MacOS and on Firefox with Ubuntu.

rien:
Oh, that is strange indeed. It seemed to be working locally.

rien:
I have not done anything and it seems to work now somehow? @mikaelGusse can you try again?

I fear there is something weird going on with the caching.

rien:
Oh, it seems to be broken when you directly load or refresh the page. Looks to be an issue with the eventlistener not being initialized.

rien:
The issue should be fixed now and I have added persistence as well.

PR #1608, opened 2024-10-08
Add support for template files into CSV parsing

mikaelGusse:
This PR makes modifications to allow for template files to be included through the info.csv metadata. Only one template file is allowed and if multiple are occurring, then an error is thrown. Basically the CSV reader now takes into account a row called 'ignored', which is then designated as the template/boilerplate code. Feel free to give feedback on this solution, and I will be happy to improve it.

Also a test was added for this feature.

### DIFF
--- a/lib/src/lib/dataset.ts
+++ b/lib/src/lib/dataset.ts
@@ -1,0 +1,161 @@
+import { readFiles, readPath } from "./reader.js";
+
+import { ExtraInfo, File, Result } from "@dodona/dolos-core";
+import { csvParse, DSVRowString } from "d3-dsv";
+
+import { constants } from "node:fs";
+import fs from "node:fs/promises";
+import path from "node:path";
+import { spawnSync as spawn } from "node:child_process";
+import { tmpdir } from "node:os";
+
+export class Dataset {
+  public nonIgnoredFiles: File[] = [];
+  public ignoredFile: File | undefined;
+  public nameCandidate: string | undefined;
+
+  constructor() {}
+
+  private async fromDirectory(dirPath: string): Promise<Result<File[]>> {
+    const dirs = [dirPath];
+    const files = [];
+
+    let i = 0;
+
+    while(i < dirs.length) {
+      for (const entry of await fs.readdir(dirs[i], { withFileTypes: true })) {
+        if (entry.isDirectory()) {
+          dirs.push(path.join(dirs[i], entry.name));
+        } else if (entry.isFile()) {
+          files.push(readPath(path.join(dirs[i], entry.name)));
+        }
+      }
+      i += 1;
+    }
+
+    return await Result.all(files);
+  }
+
+private async setIgnoredFile(resolvedFiles: File[], ignore?: string): Promise<File | undefined> {
+  const ignoredFiles = resolvedFiles.filter(file => file.extra?.ignored === "true");
+  if (ignoredFiles.length > 1) {
+    throw new Error("More than one file has the ignored field set to true. Only one template/boilerplate code file is allowed at this moment.");
+  }
+  else if (ignore) {
+      return (await readPath(ignore)).ok();
+  }
+  return ignoredFiles.length === 1 ? ignoredFiles[0] : undefined;
+}
+
+
+  private async fromZIP(zipPath: string, ignore?: string): Promise<Result<{ nonIgnoredFiles: File[], ignoredFile: File | undefined }>> {
+    const tmpDir = await fs.mkdtemp(path.join(tmpdir(), "dolos-unzip-"));
+    try {
+      const { status, error, stderr } = spawn("unzip", [zipPath, "-d", tmpDir]);
+      if (error) {
+        throw error;
+      } else if (status != 0) {
+        throw new Error(`Unzipping failed with exit status ${ status }, stderr: \n${stderr}`);
+      }
+      const infoPath = path.join(tmpDir, "info.csv");
+      if (await fs.access(infoPath, constants.R_OK).then(() => true).catch(() => false)) {
+        const { nonIgnoredFiles, ignoredFile } = (await this.fromCSV(infoPath, ignore)).ok();
+        if (nonIgnoredFiles) {
+          return Result.ok({nonIgnoredFiles, ignoredFile});
+        }
+        else {
+          throw new Error("Failed to process files");
+        }
+      } else {
+        const nonIgnoredFiles = (await this.fromDirectory(tmpDir)).ok();
+        const ignoredFile = undefined;
+        return Result.ok({nonIgnoredFiles, ignoredFile});
+      }
+    } finally {
+      await fs.rm(tmpDir, { recursive: true });
+    }
+  }
+
+
+  private async fromCSV(infoPath: string, ignore?: string): Promise<Result<{ nonIgnoredFiles: File[], ignoredFile: File | undefined }>> {
+    const dirname = path.dirname(infoPath);
+    try {
+      const files = csvParse((await fs.readFile(infoPath)).toString())
+        .map((row:  DSVRowString) => ({
+          filename: row.filename as string,
+          fullName: row.full_name as string,
+          id: row.id as string,
+          status: row.status as string,
+          submissionID: row.submission_id as string,
+          nameEN: row.name_en as string,
+          nameNL: row.name_nl as string,
+          exerciseID: row.exercise_id as string,
+          createdAt: new Date(row.created_at as string),
+          labels: row.label as string || row.labels as string,
+          ignored: row.ignored as string
+        }))
+        .map((row: ExtraInfo) => readPath(path.join(dirname, row.filename), row));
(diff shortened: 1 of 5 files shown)


lib/src/lib/dolos.ts
```diff
@@ -93,14 +107,16 @@ export class Dolos {
 
   public async analyzePaths(paths: string[], ignore?: string): Promise<Report> {
     let files = null;
+    let files_from_csv = null;
```
rien, line 110: Dolos uses mostly `camelCase` for variable names.

(I just now noticed that in the tests we are sometimes using `snake_case` by accident, apologies for the confusion :sweat_smile:)

lib/src/lib/dolos.ts
```diff
@@ -58,7 +58,13 @@ export class Dolos {
       }
       const infoPath = path.join(tmpDir, "info.csv");
       if (await fs.access(infoPath, constants.R_OK).then(() => true).catch(() => false)) {
-        return await this.fromCSV(infoPath);
+        const { nonIgnoredFiles } = (await this.fromCSV(infoPath)).ok();
```
rien, line 61: This does not seem to use the `ignored` files.

lib/src/lib/dolos.ts
```diff
@@ -81,10 +87,18 @@ export class Dolos {
...
-          labels: row.label as string || row.labels as string
+          labels: row.label as string || row.labels as string,
+          ignored: row.ignored as string
         }))
         .map((row: ExtraInfo) => readPath(path.join(dirname, row.filename), row));
-      return await Result.all(files);
+        const resolvedFiles = await Result.all(files);
+        const ignoredFiles = resolvedFiles.ok().filter(file => file.extra?.ignored === "true");
+        if (ignoredFiles.length > 1) {
+          throw new Error("More than one file has the ignored field set to true.");
```
rien, line 97:
```suggestion
          throw new Error("More than one file has the ignored field set to true. Only one template/boilerplate code file is allowed at this moment.");
```

lib/src/lib/dolos.ts
```diff
@@ -113,11 +129,19 @@ export class Dolos {
...
-    return this.analyze((await files).ok(), nameCandidate, ignoredFile);
+      if (files) {
+        if (ignore) {
+          ignoredFile = (await readPath(ignore)).ok();
+        }
+        return this.analyze((await files).ok(), nameCandidate, ignoredFile);
+      }
+      else if (files_from_csv) {
+        const { nonIgnoredFiles, ignoredFile } = (await files_from_csv).ok();
+        return this.analyze(nonIgnoredFiles, nameCandidate, ignoredFile);
```
rien, line 140: This will ignore an `ignore` file if it was given through the CLI (by passing the `--ignore` option). I think we might try supporting that as well.

If we continue with the current approach: I think it makes sense if the previous if-block (starting at line 115) would set `files`and `ignoredFile`. This way you can drop this if/else block.

rien (changes requested):
I notice this code is getting quite complicated, with many code paths doing similar things for different inputs.

Can I maybe suggest the following: creating a new `Dataset` class, which would process the input arguments (`paths`  and `ignore`) and process them similarly to how the `formCSV`, `fromZIP`, ... functions work now.

I think this would offer a few advantages:
- a `setIgnoredFile` which checks whether an ignored file was not already set and throw an appropriate error
- The `nameCandidate` logic could also be put away in a cleaner function
- we would end up with a cleaner interface to pass everything to the `analyze` function

What do you think @mikaelGusse? I am also okay with continuing the current approach.

rien:
GitHub is not running the CI workflows on your PR's, I think this is because you are making PR's from your fork?

Feel free to create your PR's in a new branch on this repository instead of your fork - this should then also run the workflows with automatic tests and so on. You should have the correct permissions to do so now.

mikaelGusse:
> I notice this code is getting quite complicated, with many code paths doing similar things for different inputs.
> 
> Can I maybe suggest the following: creating a new `Dataset` class, which would process the input arguments (`paths` and `ignore`) and process them similarly to how the `formCSV`, `fromZIP`, ... functions work now.
> 
> I think this would offer a few advantages:
> 
>     * a `setIgnoredFile` which checks whether an ignored file was not already set and throw an appropriate error
> 
>     * The `nameCandidate` logic could also be put away in a cleaner function
> 
>     * we would end up with a cleaner interface to pass everything to the `analyze` function
> 
> 
> What do you think @mikaelGusse? I am also okay with continuing the current approach.

This is probably a good idea. I will look into implementing something like this as well as looking into your other suggestions. Thanks for the feedback!

mikaelGusse:
Ok so I took a bit of a hybrid approach to this. There is now a dataset.ts which dolos.ts imports from. The Dataset class has the functions that previously existed in dolos.ts regardings the data preprocessing with some new helper functions. What do you think of this solution? I think I'll start creating these changes in this repo after this one is done to fix the CI issue. I ran lint and tests manually for this for now.

lib/src/lib/dataset.ts
```diff
@@ -0,0 +1,170 @@
...
+        }
+      }
+    }
+    else if (paths.length === 2) {
+      nameCandidate = path.basename(paths[0]) + " & " + path.basename(paths[1]);
+    }
+    return nameCandidate;
+  }
+
+  public async prepareDataset(paths: string[], ignore?: string): Promise<Dataset> {
```
rien, line 134: The `async` methods create a bit of an awkward setup where you need to construct an empty `Dataset` object first and then call this method to initialize it. I see there is indeed no way to put this part of the code in the constructor because it is async.

I suggest transforming this method to a _factory_ function like this:

```ts

class Dataset {

	constructor(
		public name: string,
		public files: File[],
		public ignore?: File) {
    }

	public static async create(paths: string[], ignore?: string): Promise<Dataset> {
    	// do async stuff
    }
}
```

Using this factory function will then be as ergonomic as:

```ts
const dataset = await Dataset.create(paths, ignore);
```

lib/src/lib/dataset.ts
```diff
@@ -0,0 +1,170 @@
...
+    let files = null;
+    let filesFromZIP = null;
+    let filesFromCSV = null;
+  
+    if (paths.length == 1) {
+      const inputFile = paths[0];
+      if (inputFile.toLowerCase().endsWith(".zip")) {
+        filesFromZIP = this.fromZIP(inputFile, ignore);
```
rien, line 142: I would immediately `await` the result of an `async`function call, as you will be using them later on anyway.

If it is wrapped in a `Result` you can do 

```ts
const { files, ignoredFile } = (await fromZIP(inputFile)).ok();
```

lib/src/lib/dataset.ts
```diff
@@ -0,0 +1,170 @@
...
+      } else {
+        throw new Error("You gave one input file, but it is not a CSV file or a ZIP archive.");
+      }
+    } else {
+      files = readFiles(paths);
+    }
+
+    this.nameCandidate = await this.createNameCandidate(paths);
+    
+    if (files) {
```
rien, line 154: The construct of checking which situation we have (ZIP-file, CSV-file or paths) occurs in multiple locations now. Maybe we can group those together in the single function for each situation?

For example:

```ts
if (inputFile.toLowerCase().endsWith(".zip") {
	return Dataset.fromZIP(inputFile, ignored);
}
```

and then the `fromZIP` function would be another static function that would create the `Dataset` instance with the `name`, `files`, etc. set.

lib/src/lib/dataset.ts
```diff
@@ -0,0 +1,170 @@
...
+      const ignoredFile = await this.setIgnoredFile(resolvedFiles.ok(), ignore);
+      const nonIgnoredFiles = resolvedFiles.ok().filter(file => file.extra?.ignored !== "true");
+      return Result.ok({ nonIgnoredFiles, ignoredFile });
+    } catch(e) {
+      throw new Error("The given '.csv'-file could not be opened");
+    }
+  }
+
+
+  public async createNameCandidate(paths: string[]): Promise<string> {
```
rien, line 117: Now that it is written out, I don't think a separate function to decide the name candidate is a good idea, because we do the same checks to see which situation we are in.

If we would need to support other checks in the future, this would require adding a case to each of these checks.

rien (commented):
This is a good improvement, but it feels a bit awkward. Can you address these comments? Feel free to propose alternatives.

mikaelGusse:
I went through your suggestions and decided to implement them. Now the functions are static such that you have this much neater await Dataset.create(...) style function call to create a Dataset. Also the name logic was moved back into their respective functions. There is less repetition now that the logic for the fromCSV and fromZIP are in their own functions that return datasets.

rien (commented):
Perfect! This looks very clean. Can you add a re-create this as a branch in this repository so the GitHub actions can run?

mikaelGusse:
I created a branch and a PR from that branch which is [here](https://github.com/dodona-edu/dolos/pull/1612)

PR #1630, opened 2024-10-22
Improve tokenizer performance

milachae:
This PR improves the tokenizer performance by using a bottom-up algorithm to walk the syntax-tree to calculate the corresponding regions for each syntax node.

On our large benchmark dataset (1100 submissions), this change brings the time spent tokenizing the files down from 25 seconds to 2.5 seconds, **making this step 10 times faster**. As more than half of the total analysis was spent tokenizing, **Dolos is now more than twice as fast**.

**API change:** The `generateTokens` method of `Tokenizer` and its inheritors `CodeTokenizer` and `CharTokenizer` now returns an array `Token[]` instead of yielding `IterableIterator<Token>`. In most cases this will not break code using this method.

### DIFF
--- a/lib/src/lib/tokenizer/charTokenizer.ts
+++ b/lib/src/lib/tokenizer/charTokenizer.ts
@@ -2,21 +2,17 @@ import { Region } from "@dodona/dolos-core";
 import { Token, Tokenizer } from "./tokenizer.js";
 
 export class CharTokenizer extends Tokenizer {
-  /**
-   * Runs the parser on a given string. Returns an async iterator returning
-   * tuples containing the stringified version of the token and the
-   * corresponding position.
-   *
-   * @param text The text string to parse
-   */
-  public *generateTokens(text: string): IterableIterator<Token> {
-    for (const [lineNumber, line] of text.split("\n").entries())
-      yield* line
-        .split("")
-        .map((char, col) => this.newToken(char, new Region(lineNumber, col, lineNumber, col + 1)));
-  }
 
-  generateTokensNew(text: string): Token[] {
-    return [...this.generateTokens(text)];
+  generateTokens(text: string): Token[] {
+    const tokens = [];
+    for (const [lineNumber, line] of text.split("\n").entries()) {
+      tokens.push(
+        line
+          .split("")
+          .map((char, col) => this.newToken(char, new Region(lineNumber, col, lineNumber, col + 1)))
+      );
+    }
+
+    return tokens.flat();
   }
 }
--- a/lib/src/lib/tokenizer/charTokenizer.ts
+++ b/lib/src/lib/tokenizer/charTokenizer.ts
@@ -15,4 +15,8 @@ export class CharTokenizer extends Tokenizer {
         .split("")
         .map((char, col) => this.newToken(char, new Region(lineNumber, col, lineNumber, col + 1)));
   }
+
+  generateTokensNew(text: string): Token[] {
+    return [...this.generateTokens(text)];
+  }
 }
--- a/lib/src/lib/tokenizer/charTokenizer.ts
+++ b/lib/src/lib/tokenizer/charTokenizer.ts
@@ -4,15 +4,14 @@ import { Token, Tokenizer } from "./tokenizer.js";
 export class CharTokenizer extends Tokenizer {
 
   generateTokens(text: string): Token[] {
-    const tokens = [];
+    const tokens: Token[] = [];
     for (const [lineNumber, line] of text.split("\n").entries()) {
-      tokens.push(
-        line
-          .split("")
-          .map((char, col) => this.newToken(char, new Region(lineNumber, col, lineNumber, col + 1)))
-      );
+      for (let col = 0; col < line.length; col++) {
+        tokens.push(this.newToken(line[col], new Region(lineNumber, col, lineNumber, col + 1)));
+
+      }
     }
 
-    return tokens.flat();
+    return tokens;
   }
 }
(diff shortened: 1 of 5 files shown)


lib/src/lib/tokenizer/charTokenizer.ts
```diff
@@ -2,17 +2,17 @@ import { Region } from "@dodona/dolos-core";
...
+    const tokens = [];
+    for (const [lineNumber, line] of text.split("\n").entries()) {
+      tokens.push(
+        line
+          .split("")
+          .map((char, col) => this.newToken(char, new Region(lineNumber, col, lineNumber, col + 1)))
+      );
+    }
+
+    return tokens.flat();
```
bmesuere, line 16: (sorry, I don't want to nitpick on code/micro manage you. I just checked this PR out of interest and saw a potential optimization)

I don't know if this code is performance critical, but it might be improved by:
- iterating over the line using a simple loop instead of `split("")` will cause a lot less overhead, no new array needs to be created, etc.
- by adding rewriting it as a loop, I think you can add the stuff directly to tokens, avoiding `flat`

There might of course be reasons why the code was written like this that I'm not aware of.

bmesuere (changes requested):

lib/src/lib/tokenizer/charTokenizer.ts
milachae: > I don't know if this code is performance critical

This code will only be used when you try to tokenize a `.txt` or `.md` file. Thus I did not really look at ways to improve this code and just adjusted it to return a list instead of an iterator. This to meet the return type specified by the abstract class.

But my introduced `.flat()` is indeed totally not necessary here. I will change this!

lib/src/lib/tokenizer/tokenizer.ts
```diff
@@ -36,7 +36,7 @@ export abstract class Tokenizer {
    * @param text The buffer to stringify
    */
   public tokenize(text: string): string {
-    return Array.of(...this.generateTokens(text)).join();
+    return Array.of(this.generateTokens(text)).join();
```
rien, line 39:
```suggestion
    return this.generateTokens(text).join();
```

`generateTokens` already returns an array, so no need to make a new one.

lib/src/lib/tokenizer/codeTokenizer.ts
```diff
@@ -32,56 +32,54 @@ export class CodeTokenizer extends Tokenizer {
...
-    yield this.newToken(node.type, location);
+      // If the code is already captured in one of the children, the region of the current node can be shortened.
+      if ((childStartRow < location.endRow) || (childStartRow === location.endRow && childStartCol < location.endCol)) {
+        location.endRow = childStartRow;
+        location.endCol = childStartCol;
+      }
 
-    for (const child of node.namedChildren) {
-      yield* this.tokenizeNode(child);
+      allNodes.push(...childNodes);
```
rien, line 77: This spread operator is probably applied on large lists of `childNodes`, which could have a large performance impact.

In addition, this also copies the items previously added to that list to another list. Passing the tokens then actually becomes a $$\Theta(n^2)$$ operation as some tokens can be moved multiple times.

I suggest passing the array to add to as an argument instead of returning it. This also has the added benefit of allowing a cleaner return type:

```ts
public generateTokens(text: string): Token[] {
    const tree = this.parser.parse(text, undefined, { bufferSize: Math.max(32 * 1024, text.length * 2) });
    const tokens = [];
	tokenizeNode(tree.rootNode, tokens);
	return tokens;
  }

private tokenizeNode(node: SyntaxNode, tokens: Token[]): [number, number] {
	const location = new Region(node.startPosition.row, node.startPosition.column, node.endPosition.row, node.endPosition.column);
	tokens.push(this.newToken("(", location);
	tokens.push(this.newToken(node.type, location);
	for (const child of node.namedChildren) {
		const [childStartRow, childStartCol] = this.tokenizeNode(child, tokens);
		// If the code is already captured in one of the children, the region of the current node can be shortened.
		if ((childStartRow < location.endRow) || (childStartRow === location.endRow && childStartCol < location.endCol)) {
			location.endRow = childStartRow;
			location.endCol = childStartCol;
		}
	}
	tokens.push(this.newToken(")", location);
	return [location.startRow, location.startCol];
}
```

lib/src/lib/tokenizer/tokenizer.ts
```diff
@@ -49,10 +49,10 @@ export abstract class Tokenizer {
   public tokenizeWithMapping(text: string): [string[], Region[]] {
     const resultTokens: Array<string> = [];
     const positionMapping: Array<Region> = [];
-    for (const { token, location } of this.generateTokens(text)) {
+    this.generateTokens(text).forEach(({ location, token }) => {
       resultTokens.push(token);
```
rien, line 53: I personally prefer keeping the `for` loop instead of a `forEach`. This is consistent with the other loops we are using in the `lib` and `core` modules of Dolos.

bmesuere: Not sure if it matters here, but in addition, forEach is up to 3 times slower compared to a simple for.

lib/src/lib/tokenizer/codeTokenizer.ts
milachae: This was indeed something I already tested last week, but (surprisingly) didn't have any impact on the execution time.

Although, it indeed does make the return type more clear.

rien:
Nevermind my suggestion to recreate the branch, it somehow started doing the CI on this PR.

rien (approved):

milachae:
> Nevermind my suggestion to recreate the branch, it somehow started doing the CI on this PR.

I think this happened because I just created the new branch in this repo. But I don't know why that triggered the CI in this PR?

PR #1643, opened 2024-11-14
Add template code highlighting

mikaelGusse:
This PR adds the ability for the user interface to highlight code fragments that have been ignored in the comparison. The ignored parts are highlighted with a faint grey color to separate from the blue that is used for matching fragments. Feel free to suggest any improvements to the implementation or functionality!

### DIFF
--- a/web/src/api/models/pair.ts
+++ b/web/src/api/models/pair.ts
@@ -1,4 +1,5 @@
 import { File, Fragment } from "@/api/models";
+import { Region } from "@dodona/dolos-core";
 
 export interface Pair {
   id: number;
@@ -10,4 +11,6 @@ export interface Pair {
   leftCovered: number;
   rightCovered: number;
   fragments: Fragment[] | null;
+  leftIgnoredKgrams: Region[];
+  rightIgnoredKgrams: Region[];
 }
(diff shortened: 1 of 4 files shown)


rien (commented):
Code is looking good. However, I'm not convinced of the color of the highlight currently:

### Current situation `rgba(0, 0, 0, 0.05)`
image

Since it is completely black with only 5% opacity, it looks blue-ish somehow - which is very similar to the match highlight color. One could mistakenly think this is just a "faint" match.

### Suggested change `rgba(220, 220, 220, 1)`

By making the ignored highlight color opaque, it is better distinguishable from the background and also "overwrites" the match highlight color if it overlaps.

image

However, in both situations I don't know if it is clear for users who do not know this feature whether we are highlighting ignored code or not. @bmesuere what is your opinion on this?

mikaelGusse:
I agree that this new color is certainly clearer. I was also wondering if there should be some "legend" which could explain what the different colors mean. It could be behind a question mark on the page somewhere perhaps? Or behind some hoverable element? Not sure about that

rien:
I have been thinking about this issue, but cannot find a good satisfying solution.

To avoid stalling this PR I will approve & merge it and create a new issue to discuss these options further.

bmesuere:
@rien sorry, I missed the mention somehow. Since the code is ignored, I would also try to (also) "fade" the code itself instead of only the background. Note that we will probably need a similar solution if/when we ignore comments.

PR #1645, opened 2024-11-19
Exclude the comments in the tokenization step

milachae:
By default, comments will now be excluded during the tokenization of a file.

I also added a CLI option (`-ic` or `--include-comments`) to include the comments during tokenization.


cli/src/cli/commands/run.ts
```diff
@@ -166,6 +166,12 @@ export function runCommand(program: Command): Command {
       x => parseFloat(x),
       Options.defaultKgramsInWindow
     )
+    .option(
+      "-ic, --include-comments",
```
rien, line 170:
```suggestion
      "-C, --include-comments",
```

The flag `-ic` feels like a combination of `-i`  and `-c` for me, and might be confusing.

lib/src/lib/tokenizer/tokenizer.ts
```diff
@@ -6,9 +6,19 @@ export interface Token {
...
 export abstract class Tokenizer {
 
-  constructor(public readonly language: Language) {}
+  protected options: TokenizerOptions = {};
+
+  constructor(public readonly language: Language, options?: TokenizerOptions) {
+    if (options !== undefined) {
+      this.options = options;
+    }
+  }
```
rien, line 21:
```suggestion
  constructor(public readonly language: Language, options: TokenizerOptions = {}) {}
```

lib/src/test/tokenizer.test.ts
```diff
@@ -27,26 +27,26 @@ const languageFiles = {
...
   "../samples/scala/Caesar.scala": 366,
-  "../samples/sql/sample.sql": 543,
-  "../samples/tsx/sample.tsx": 1659,
-  "../samples/typescript/caesar.ts": 378,
-  "../samples/verilog/module.v": 2484
+  "../samples/sql/sample.sql": 540,
+  "../samples/tsx/sample.tsx": 1656,
+  "../samples/typescript/caesar.ts": 375,
+  "../samples/verilog/module.v": 2448
 } as {[key: string]: number};
```
rien, line 50: What was the purpose of this particular test (checking the token length) in addition to the tokenizer snapshots?

You've added this in a previous PR, but I see this will change often when the tokenizers change and might me extra manual work - whereas the snapshots can be automatically updated if they change.

rien (changes requested):
Can you add a test to check if setting `includeComments = true` actually includes the comments?

lib/src/test/tokenizer.test.ts
milachae: I made this test to check if my changes to the tokenizer did change the output.  but at that time, I didn't know what the snapshots did. So this test is indeed redundant and I removed it.

rien (approved):

PR #1646, opened 2024-12-02
Add support for OCaml programming language

rien:
This PR adds support for the OCaml programming language (https://ocaml.org/).


rien:
@kudhru  Usually installing `tree-sitter-ocaml` would be enough for the Dolos CLI to detect this parser locally, but this is not possible due to this nonstandard structure of the parser.

So I've added support for OCaml in this PR, you should be able to run Dolos locally with this code.

We were planning on making a new release soon. I'll do my best to do this before the end of this week, but I cannot promise anything unfortunately. Feel free to let me know if you have any issues getting the CLI working locally.

kudhru:
Thanks, will give it a try!

PR #1656, opened 2025-01-23
fixed a typo in the CLI help message

eriq-augustine:
The CLI help prompt used "chars" instead of "char" as the character-by-character comparison.


rien (approved):
This is indeed an error! Thanks for spotting it and submitting a fix.

PR #1660, opened 2025-03-03
Uncomment pagination for clusters

mikaelGusse:
Hello!

We noticed in our own usage of Dolos that the clusters are missing pagination. One of us found out that it is simply commented out in the code. She tried commenting it out and didn't notice that anything broke. I also tried it and it seemed ok. Was there some good reason for this page to not be paginated that we missed?

This PR basically just takes out those lines of code that disable pagination and adds a couple of rows to the page.

Feel free to give feedback if there is some issue with this

Thanks!

### DIFF
--- a/web/src/components/clustering/ClustersTable.vue
+++ b/web/src/components/clustering/ClustersTable.vue
@@ -4,7 +4,7 @@
     :headers="headers"
     :items="items"
     :sort-by="sortBy"
-    :items-per-page="15"
+    :items-per-page="25"
     must-sort
     fixed-header
     @click:row="rowClicked"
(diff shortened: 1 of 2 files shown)


rien:
Hi @mikaelGusse we disabled this functionality because we didn't expect a lot of clusters so that pagination is needed. But now I notice that instead of having just a single list, it simply hides clusters if there are more than 15.

So this is definitely a good fix, thank you for bringing it up.

rien:
Upon closer inspection, there seemed to be something else going wrong here.

The clustering table is not meant to have pagination buttons on the overview page, because it should bring the user to the _View by clusters_ page if they want to see more. Thus the conditional check was added to disable it on demand.

Now either we forgot to check the clusters page afterwards, and didn't set a correct value for the `pagination` prop. I've added this now so that pagination is enabled by default. I've also clarified the behavior in the corresponding comment.

@mikaelGusse I've set the `items-per-page` back to 15 because 25 items are too much for a 1080p screen (so it shows a scrollbar).  What do you think?

mikaelGusse:
Ok that makes sense, thanks for delving further into this. Yes this newest version is very much OK in my opinion.

PR #1665, opened 2025-03-17
Improve web container build process

FWest98:
This MR restructures the Docker build process for the `web` image to improve caching and reusability of the layers:
- We move the installation of `curl` to the very front, so it is always reused. 
- We separately copy `package.json` and install the dependencies first, before copying all other files.

This way, when only some web code changes, it is no longer needed to install all dependencies again.


rien (approved):
This is a good improvement, thank you for your contribution!

PR #1666, opened 2025-03-17
Improve Docker Compose setup flexibility through `.env` file

FWest98:
This MR proposes some improvements to the preprovided Docker Compose setup. In particular:
- Easily enable user customisation through an `.env` file
- Allow customisation of database credentials
- Restructure ports and container environment variables to now use a single set of hosting parameters, that allows for both local and public hosting behind a reverse proxy.
- Include relevant documentation

My main motivation of including these changes is to simplify hosting Dolos behind a reverse proxy. I spent quite some time getting everything to work properly in a "production"-ready setup, due to the complex interplay between container-specific variables and ports. This of course does not include the configuration of a reverse proxy itself, which should still be done by the user.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,31 @@
+# The protocol on which Dolos will be accessed by the user - HTTP or HTTPS
+WEB_PROTOCOL=http
+
+# Database configuration
+DATABASE_ROOT_PASSWORD=root
+DATABASE_USER=dolos
+DATABASE_PASSWORD=dolos
+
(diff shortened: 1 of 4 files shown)


rien (approved):
Hi @FWest98, this is a wonderful improvement! Thanks for updating the documentation as well. :pray: 

It seems like you're setting up your own instance, if there is anything we can do to help or if you have any feedback, you're always welcome to contact us.

PR #1849, opened 2026-04-30
Make dolos-parsers an optional dependency in lib

### LINKED ISSUE #1839: Create a slimmer version of lib package
**Is your feature request related to a problem? Please describe.**

As I was starting to play around with the library, I noticed that it has a pretty big footprint in the form of a dependency on `@dodona/dolos-parsers`. In my environment that package alone consumes around 435MB. However, if my understanding of the content of the package is correct, it seems most of these parsers are content that I'll typically never use, since I don't need the parsers for OCAML (98MB), Verilog (46MB), etc. The CLI has the same issue, since it depends on the lib. Since I'm hoping to integrate this library into other projects that involve building Docker containers, having a small footprint probably makes sense.

**Describe the solution you'd like**

Ideally I'd like to have a way to "pick and choose" which parsers I'm interested in, and only install the ones that I'll actually use. So the library would install only the bare minimum parser functionality, and then if I need, for example, Python and C, I would install the parsers for those languages (could be something like `@dodona/dolos-parser-python`, for example). This would mean that the lib/cli would fail if I try to use a language that is not installed, and that is fine.

This would also provide additional flexibility for users to add their own parsers outside of the dolos environment.

**Describe alternatives you've considered**
NA

**Additional context**
Other libraries that use the functionality above with optional packages include:
* Highlight.js languages: https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md
* Mathjax fonts: https://docs.mathjax.org/en/v4.1/output/fonts.html

milachae:
This PR removes `"@dodona/dolos-parsers": "1.4.1"` as dependency in `@dodona/dolos-lib`. The package can still be added by the user.

If a parser for a specific language is not available in `@dodona/dolos-parsers`, or if the package is not installed, the library will automatically fall back to using the corresponding `tree-sitter-<language>` package instead.

### DIFF
--- a/docs/docs/library.md
+++ b/docs/docs/library.md
@@ -13,6 +13,13 @@ Using `npm` to install the library, similar to [installing the CLI](/docs/instal
 npm install @dodona/dolos-lib
 ```
 
+By default, the library uses parsers provided by `@dodona/dolos-parsers`. You can install them with:
+```
+npm install @dodona/dolos-parsers
+```
+
+If a parser for a specific language is not available in `@dodona/dolos-parsers`, or if the package is not installed, the library will automatically fall back to using the corresponding `tree-sitter-<language>` package instead.
+
 ## Usage
 
 Take a look at our [example repository](https://github.com/rien/dolos-lib-example/blob/main/index.mjs)
--- a/lib/README.md
+++ b/lib/README.md
@@ -11,7 +11,7 @@ Visit [dolos.ugent.be](https://dolos.ugent.be) for more information.
 npm install @dodona/dolos-lib
 ```
 
-By default, the library uses parsers provided by `@dodona/dolos-parsers`. You can install them with:
+By default, the library tries to use the parsers provided by `@dodona/dolos-parsers`, which must be installed separately:
 ```
 npm install @dodona/dolos-parsers
 ```
--- a/lib/README.md
+++ b/lib/README.md
@@ -11,6 +11,13 @@ Visit [dolos.ugent.be](https://dolos.ugent.be) for more information.
 npm install @dodona/dolos-lib
 ```
 
+By default, the library uses parsers provided by `@dodona/dolos-parsers`. You can install them with:
+```
+npm install @dodona/dolos-parsers
+```
+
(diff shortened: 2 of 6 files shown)


docs/docs/library.md
```diff
@@ -13,6 +13,13 @@ Using `npm` to install the library, similar to [installing the CLI](/docs/instal
 npm install @dodona/dolos-lib
 ```
 
+By default, the library uses parsers provided by `@dodona/dolos-parsers`. You can install them with:
```
rien, line 16:
```suggestion
This library uses tree-sitter parsers to process source code files. We've bundled a large amount of parsers in one package `@dodona/dolos-parsers` for convenience. Because this package is quite large (+400MB), we don't depend on this package. However, you can install this package with:
```

docs/docs/library.md
```diff
@@ -13,6 +13,13 @@ Using `npm` to install the library, similar to [installing the CLI](/docs/instal
...
+```
+npm install @dodona/dolos-parsers
+```
+
+If a parser for a specific language is not available in `@dodona/dolos-parsers`, or if the package is not installed, the library will automatically fall back to using the corresponding `tree-sitter-<language>` package instead.
```
rien, line 21:
```suggestion
This library will automatically fall back on the corresponding `tree-sitter-<language>` parser instead, so if you're using this library only for a select number of programming languages you can simply install those tree-sitter parsers directly and the library will automatically detect their presence.
```

lib/src/lib/language.ts
rien, line 54: Consider adding a suggestion here to install `@dodona/dolos-parsers` if it is not installed?

lib/README.md
```diff
@@ -11,6 +11,13 @@ Visit [dolos.ugent.be](https://dolos.ugent.be) for more information.
 npm install @dodona/dolos-lib
 ```
 
+By default, the library tries to use the parsers provided by `@dodona/dolos-parsers`, which must be installed separately:
+```
+npm install @dodona/dolos-parsers
+```
+
+If a parser for a specific language is not available in `@dodona/dolos-parsers`, or if the package is not installed, the library will automatically fall back to using the corresponding `tree-sitter-<language>` package instead.
```
rien, line 19: Update this according to the changed docs, or link to that section to avoid duplication?

rien (commented):
Should we add tests to check whether this works?

PR #1877, opened 2026-06-23
Add frontend test suite (unit + visual regression)

rien:
Introduces a frontend test safety net for the `web` package, which previously had no tests.

### Unit tests (Vitest)
- Vitest config (`vitest.config.ts`, happy-dom), typed fixtures (`src/test/fixtures.ts`), and `tsconfig` excludes so the production `vue-tsc` build ignores tests.
- 39 tests covering pure-logic seams: sorting, `PairAsNestedMap`, similarity cutoff/threshold, clustering algorithms, file interestingness, and CSV cast/file utils.

### Visual-regression tests (Playwright)
- Full-app harness (`tests/visual/routes.spec.ts`) that boots the real Vite dev server, loads the real report from `public/data`, and screenshots every analysis route against committed baselines via `toHaveScreenshot`.
- A run **fails automatically** on visual drift; `npx playwright show-report` shows an Expected/Actual/Diff slider for human review.

### Graph determinism
- The D3 force graph caused these visual checks to fail because of non-determinism. This PR made the simulation more deterministic by using a seeded random number generator and by creating a new function  `seedClusterLayout` (`graph/layout.ts`) that packs clusters largest-first with `d3.packSiblings`, filling each with a deterministic golden-angle disk, so the simulation converges to a single steady state.
- Seeding runs only on the **initial** render; later updates reuse the existing layout.

## Notes
- Storybook scaffolding was removed in favour of the Playwright full-app approach.
- NixOS integration via `shell.nix` (`playwright-driver.browsers` + `PLAYWRIGHT_BROWSERS_PATH`), with `@playwright/test` pinned to the nixpkgs driver version.
- Baseline PNGs are committed (they are the references); regenerate intentionally with `npm run test:visual -- --update-snapshots`.


.github/workflows/ci-cd.yml
```diff
@@ -471,6 +471,96 @@ jobs:
...
+        run: |
+          cd web/
+          npm run test
+
+  web-visual-test:
+    name: "Web 🌐 - visual tests 📸"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v6
```
milachae, line 522: This was recently updated to `v7` for the other ci/cd

milachae (approved):
Great PR! Thanks for adding these visual tests, they will definitely be useful.

milachae (approved):

PR #1889, opened 2026-06-25
Remove dead code and unused exports, fix memory leaks

rien:
Frontend cleanup and store-layer restructure in the `web/` package.

### Cleanup
- Removed dead code (unused files and symbols) and narrowed over-broad exports.
- Fixed memory/DOM leaks in the D3 graph and tooltip composables.

### Store-layer restructure
- Consolidated all Pinia stores under `src/stores/` (with a `report/` subfolder for the open report); `src/api/` is now pure data-layer code (parsers, worker, models, utils) with no Pinia.
- Replaced the misnamed `api.store` with a `useReportLoader` composable that orchestrates hydration and owns `loading`/`loadingText`/`error`.
- Extracted `cutoff*` and `isAnonymous` into a shared `settings.store`; data stores now read their CSV URL from `useAppMode().dataUrl` directly.
- Extracted CSV parsing into pure `parseFiles`/`parsePairs` functions (with unit tests).
- Restored the loading overlay during anonymize and fixed an anonymisation loading bug.

### Verification
`npm run check`, `npm test`, and the Playwright visual harness (`npm run test:visual`).


web/src/api/stores/file.store.ts
```diff
@@ -275,7 +275,6 @@ export const useFileStore = defineStore("file", () => {
       filesById.value = { ...filesById.value };
       pairStore.pairs = { ...pairStore.pairs };
       legend.value = { ...legend.value };
-      apiStore.loading = false;
```
milachae, line 278: anonymize() no longer sets loading, so the "Anonymizing files..." overlay is gone. Could this be a problem for reports with many files?

web/src/stores/report/metadata.store.ts
```diff
@@ -1,44 +1,44 @@
...
+    return Object.fromEntries(
+      data.map(row => [row.property, castToType(row).value])
+    );
+  }
+
+  // Reference to other stores.
+
+  // Fetch the metadata from the CSV file.
+  async function fetch(): Promise<any[]> {
+    const { dataUrl } = useAppMode();
```
milachae: Is there a reason to move the `useAppMode()` into the `fetch` function? This is done in `metadata.store.ts`  and `kgram.store.ts` but not in the pair and file stores.

milachae (commented):
This cleans up the frontend a lot! I just have 2 small questions.

milachae (approved):

PR #1890, opened 2026-06-25
Continued refactor: restructure Pinia stores

rien:
This also includes a newline substitution (somehow there were lots of carriage returns in most files).


rien:
The work from this PR is included in a previous PR (#1889)

PR #1894, opened 2026-06-30
Cleanup and simplify custom CSS

rien:
Throughout the years we've had to add a lot of custom CSS. Primarily due to Vue(tify) upgrades that had some lacking styling options. However, these styling issues should be resolved now, but the custom CSS remained.

The visual checks using playwright didn't contain a test for the upload page, so we first added those with a stubbed API endpoint.

The rest of this PR attempts to simplify and cleanup this custom CSS.

In addition, we've also fixed a visual issue on the overview page where the labels table had a fixed height that was sometimes smaller than its container. This table should now grow in a more visually pleasing way.


milachae (approved):
Thank you for cleaning things up!

